PR #1, opened 2026-04-08
refactor: move argument parsing to separate module

Gipphe:


ghcib/src/Ghcib/Arguments.hs
cgeorgii, line 1: Could we name this file something more appropriate, like Cli? Arguments is not great since we have commands, etc.

cgeorgii (approved):
Other than file naming, LGTM.

ghcib/src/Ghcib/Arguments.hs
Gipphe: I was thinking in terms of "CLI arguments". "CLI" sounds more like the actual CLI itself rather than the arguments you pass to the CLI program.

PR #3, opened 2026-04-08
feat(ghcib): add module source lookup

cgeorgii:
- New `ghcib source MODULE...` command prints Haskell source for installed modules via Haddock hyperlinked-source HTML
- `GhcPkg` effect wraps `ghc-pkg find-module` / `ghc-pkg field`
- `lookupModuleSource` resolves module → package → haddock path → source
- Two-level cache: `ModuleName → PackageId` and `(PackageId, ModuleName) → Text`
- `Cache.runCacheForever`: lightweight no-TTL interpreter (useful for tests and session-scoped caches)
- `FileSystem(..)` constructors exported to support external interpreters
- Tests: `SourceSpec` (pure HTML extraction), `GhcPkgSpec` (scripted interpreter), `SourceLookupSpec` (end-to-end with stubs)


ghcib/src/Ghcib/Effects/GhcPkg.hs
```diff
@@ -0,0 +1,68 @@
...
+    GetHaddockHtml :: PackageId -> GhcPkg m (Maybe FilePath)
+
+
+makeEffect ''GhcPkg
+
+
+runGhcPkgIO :: (IOE :> es) => Eff (GhcPkg : es) a -> Eff es a
+runGhcPkgIO = interpret \_ -> \case
+    FindModule modName -> do
+        out <- readProcessSafe "ghc-pkg" ["find-module", "--simple-output", toString (unModuleName modName)]
```
Gipphe, line 34: question: I take it `ghc-pkg` doesn't have a library we can use instead of shelling out?

Gipphe (approved):
I gave this a cursory read, but this is very much a rubber stamp approval 😅 But that's fine for now either way 😛

ghcib/src/Ghcib/Effects/GhcPkg.hs
cgeorgii: For now I would like to avoid getting coupled to a GHC version, instead using whatever is available from the user's environment.

PR #7, opened 2026-04-13
refactor: use mockable Console effect

Gipphe:
Replaces the use of `effectful`'s `Effectful.Console.ByteString` effect with a dynamic effect we can use to more easily test code that writes to the console (which this application does a lot).


cgeorgii (approved):
Nice, thanks :)

PR #8, opened 2026-04-13
refactor: use File effect for operations on file handles

Gipphe:
Depends on #7


atelier/src/Atelier/Effects/File.hs
```diff
@@ -0,0 +1,81 @@
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
```
cgeorgii, line 1: I assume this is necessary?

cgeorgii (approved):

PR #14, opened 2026-04-15
feat(ghcib): expose flake outputs for external consumption

cgeorgii:
- `overlays.default` adds `pkgs.ghcib` to nixpkgs
- `homeManagerModules.default` — `programs.ghcib.enable` via Home Manager
- `nixosModules.default` — `programs.ghcib.enable` for non-HM NixOS users
- `apps.ghcib` enables `nix run github:atelier-hub/atelier#ghcib`
- `packages.ghcib` / `packages.default` wrap `ghcib-exe` so the binary is named `ghcib`
- `checks.overlay` validates the overlay end-to-end in CI
- Pass `self` into `outputs.nix` to support the overlay check derivation
- Document `nix run`, dev shell, and installation options in README

### DIFF
--- a/nix/home-module.nix
+++ b/nix/home-module.nix
@@ -1,0 +1,19 @@
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
+let
+  cfg = config.programs.ghcib;
+in
+{
+  options.programs.ghcib = {
+    enable = lib.mkEnableOption "ghcib GHCi build daemon";
+    package = lib.mkPackageOption pkgs "ghcib" { };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+  };
+}
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -1,0 +1,19 @@
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
+let
+  cfg = config.programs.ghcib;
+in
+{
+  options.programs.ghcib = {
+    enable = lib.mkEnableOption "ghcib GHCi build daemon";
+    package = lib.mkPackageOption pkgs "ghcib" { };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+  };
+}
--- a/nix/outputs.nix
+++ b/nix/outputs.nix
@@ -1,4 +1,8 @@
-{ inputs, system }:
+{
+  inputs,
+  system,
+  self,
+}:
 let
   # GHC version to use across all tools and the project
   compiler-nix-name = "ghc9102";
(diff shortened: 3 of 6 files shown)


nix/home-module.nix
```diff
@@ -0,0 +1,18 @@
...
+  lib,
+  config,
+  ...
+}:
+let
+  cfg = config.programs.ghcib;
+in
+{
+  options.programs.ghcib = {
+    enable = lib.mkEnableOption "ghcib GHCi build daemon";
```
Gipphe, line 12: suggestion: Add a `package` option.

```suggestion
    enable = lib.mkEnableOption "ghcib GHCi build daemon";
    package = lib.mkPackageOption pkgs "ghcib" {};
```

nix/home-module.nix
```diff
@@ -0,0 +1,18 @@
...
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ pkgs.ghcib ];
```
Gipphe:
```suggestion
    home.packages = [ cfg.package ];
```

nix/nixos-module.nix
```diff
@@ -0,0 +1,18 @@
...
+  lib,
+  config,
+  ...
+}:
+let
+  cfg = config.programs.ghcib;
+in
+{
+  options.programs.ghcib = {
+    enable = lib.mkEnableOption "ghcib GHCi build daemon";
```
Gipphe, line 12:
```suggestion
    enable = lib.mkEnableOption "ghcib GHCi build daemon";
    package = lib.mkPackageOption pkgs "ghcib" {};
```

nix/nixos-module.nix
```diff
@@ -0,0 +1,18 @@
...
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [ pkgs.ghcib ];
```
Gipphe:
```suggestion
    environment.systemPackages = [ cfg.package ];
```

nix/outputs.nix
```diff
@@ -83,7 +94,8 @@ in
 {
   # Expose packages built by haskell.nix
   packages = projectFlake.packages // {
-    default = projectFlake.packages."atelier:exe:ghcib-exe";
+    default = ghcib;
```
Gipphe, line 97: thought: It could be argued that this repo doesn't have a proper "default" package, since it is named "atelier" and contains the "atelier" library.

Gipphe (approved):
Apart from the package options, this looks good to me! Up to you whether you think it's worth it to supply the package options. Might be unnecessary at this point since we don't have any other options that would use the package, so 🤷

nix/outputs.nix
cgeorgii: We can remove the package once we extract ghcib from it.

PR #20, opened 2026-04-17
feat: use brick for TUI

Gipphe:
This PR merely introduces `brick`, while attempting to keep things as similar to the existing setup as possible.

Old view:
image

New view:
image


cgeorgii (approved):
Awesome, I like it!

PR #21, opened 2026-04-20
fix: restart GHCi session on cabal file changes

### LINKED ISSUE #73: Failing to watch changes on a multi-package repo
## Problem

I trying tricorder on this project branch: https://github.com/tweag/Diff/tree/ninioArtillero/static-checks using `nix run --accept-flake-config github:atelier-hub/tricorder -- ui` wth the following configuration:

```yaml
#.tricorder.yaml
session:
  command: "cabal repl Diff-liquidhaskell:test:check --repl-options='-fno-break-on-exception -fno-break-on-error -v1 -ferror-spans -j'"
  targets:
    - "Diff-liquidhaskell:test:check"
  test_targets: []
  watch_dirs:
    - src
    - Diff-liquidhaskell/src
```

And get this output

```
All good. (1 modules, 1.4s)   10:51:37

Targets: Diff-liquidhaskell:test:check
Watching:
  - ./src
  - ./Diff-liquidhaskell/src
Socket: /run/user/1000/tricorder/4946307a8da4de73/socket.sock
Metrics: disabled

Press 'h' for help
```

Somehow the build succeeded, even though I purposefuly currupted the source by adding values that should not typecheck (both for Haskell and LiquidHaskell perspective). Further (bad) changes are noticed, don't affect the `All good` status.

Furthermore, when run without the configuration file, the main package is picked (even though I disabled this package by removing it from `cabal.project`, but I understand this is not supported), but the build hangs:

```
Building...                                                                                                    
                                                                                                                                                                                                                                                                           Targets: lib:Diff test:diff-tests
 Watching:
 - ./src
 - ./test
 Socket: /run/user/1000/tricorder/4946307a8da4de73/socket.sock
Metrics: disabled
```

With just the session command configured as before, the build succeeds and the daemon notices the changes as before, but again does not fail when it should.

Note: I did both `tricorder stop` and `pkill tricorder` between experiments.

## Context

Just some context on the project in the hope that it is relevant to the issue.

## Package split

The project has two packages:
- `Diff`, providing the main library under `./src` and a test suite.
- `./Diff-liquidhaskell`, a sub project with a single test suite co2mponent sharing sources with the `Diff` library:

```
  test-suite check
  type:             exitcode-stdio-1.0
  default-language: Haskell2010
  main-is:          Main.hs
  -- Re-use the main Diff source tree so LH checks the real code.
  hs-source-dirs:   src, ../src
  other-modules:
    Data.Algorithm.Diff
    Data.Algorithm.DiffOutput
    Data.Algorithm.DiffContext
    Internal.LiftedFunctions
  -- Same as Diff, but adding liquidhaskell.
  build-depends:
      base          >= 4.11 && <= 6
    , array
    , pretty        >= 1.1
    , liquidhaskell
    , liquid-prelude
  ghc-options: -Wall -O0 -fplugin=LiquidHaskell -fplugin-opts=LiquidHaskell:"--dump-opaque-reflections" -fforce-recomp
```

LiquidHaskell `check` depends on a dummy `Main` module that just imports all `Data.Algorithm` modules, so that their build has the LH pluging enabled. We force recompilation to make sure updated specs are always checked, even when only comments where edited (LH specs are just comments to cabal).

I usually run the test by `cabal run Diff-liquidhaskell:test:check` when debugging a spec, or use

`ghcid "--command=cabal repl Diff-liquidhaskell:test:check --repl-options='-fno-break-on-exception -fno-break-on-error -v1 -ferror-spans -j'"`

when tunning a spec (this is because `ghcid` swallows big chunks of the error messages).

## `cabal.project` tweaks
Furthermore, I have the following `cabal.project`, so that I can build agains latest LH:

```
-- GHC 9.14 compatibility relaxations (from LH's own cabal.project).
allow-newer:
    aeson:containers
  , aeson:template-haskell
  , aeson:time
  , indexed-traversable:base
  , indexed-traversable:containers
  , indexed-traversable-instances:base
  , lens-family:containers
  , lens-family-core:containers
  , semialign:base
  , semialign:containers
  , text-iso8601:time
  , time-compat:time
  , these:base
  , uuid-types:template-haskell

constraints:
  aeson >= 2.2.3.0
```

And use `cabal.project.local` to point to my local checkout:

```
with-compiler: ghc-9.14.1

packages:
   ./Diff-liquidhaskell
   ../liquidhaskell
   ../liquidhaskell/liquidhaskell-boot
   ../liquidhaskell/liquid-prelude
   ../liquidhaskell/liquid-fixpoint
```

cgeorgii:
`tricorder` failed to restart the GHCi session when cabal-related files changed (e.g. adding a `build-depends` entry). This PR fixes that with two complementary changes and promotes both effects to `atelier` as reusable primitives.
## Changes

**FileWatcher** (`Atelier.Effects.FileWatcher`)
- Replaces the old `WatchDirs :: [FilePath]` with a `WatchFilePaths :: [Watch]` DSL — `dir`, `dirWhere`, `dirExt`, `excluding`, `containing` — giving callers fine-grained control over which files trigger callbacks
- Canonicalizes Watch paths to absolute before calling `matchesAny`, fixing a silent mismatch against fsnotify's always-absolute event paths
- Deduplicates watched directories by path prefix to register the minimal set of OS-level watchers
- Watcher now explicitly watches `*.cabal`, `cabal.project`, and `package.yaml` via a predicate on `projectRoot`, so cabal changes are detected and routed to `RestartSession`

In practice the watch list reads like a small spec:

```haskell
sourceWatches :: [FilePath] -> [Watch]
sourceWatches dirs = map (\d -> dirExt d ".hs" `excluding` containing "dist-newstyle") dirs

cabalWatches :: FilePath -> [Watch]
cabalWatches projectRoot = [dirWhere projectRoot isCabalFile]
```

And the watcher callback ties both together:

```haskell
watchFilePaths (sourceWatches dirs <> cabalWatches projectRoot) \path ->
    debounced path (markDirty (changeKindFor path))
```

**Debounce** (`Atelier.Effects.Debounce`)
- Replaces the channel-based `Tricorder.Debounce` with a keyed STM/fork implementation — `debounced key action` schedules `action` after a settle window, cancelling it if another event for the same key arrives first
- Applied in the file watcher callback (`debounced path (markDirty ...)`) to coalesce rapid inotify bursts (e.g. two events for the same cabal file) into a single `markDirty` call
- `runDebounce` reads `debounceMs` from `Config` and is wired into the main effect stack

Both effects and their tests now live in `atelier`/`atelier-test`.


atelier/src/Atelier/Effects/Debounce.hs
```diff
@@ -0,0 +1,88 @@
...
+    => Eff (Debounce key : es) a
+    -> Eff es a
+runDebounce eff = do
+    counters <- STM.atomically (Map.new :: STM (Map.Map key Int))
+    interpretWith eff \env -> \case
+        Debounced settleMs key action ->
+            localUnlift env concStrat \unlift -> do
+                n <- STM.atomically $ do
+                    current <- Map.lookup key counters
+                    let n = maybe 0 (+ 1) current
```
Gipphe, line 70: thought: A bit besides the point in this case, but `maybe 0` means that the first instance to hit the debounce will have a count of `0`, while the second one will increment the count to `1`, etc. Should it start on `1`?

Not a crucial detail in this case really, since it's just used to ensure only the last call to `debounced` is actually fired off in the end.

atelier/src/Atelier/Effects/Debounce.hs
```diff
@@ -0,0 +1,88 @@
...
+                    Map.insert n key counters
+                    pure n
+                void $ fork $ do
```
Gipphe, line 73: question (non-blocking): This might potentially fire off a bunch of virtual threads if called in rapid succession.

```hs
[1..1_000_000] <&> \i -> debounced 300 () (putStrLn $ show i)
```

The above snippet will spawn 1 million threads, which will probably take up a significant amount of memory. Granted, we probably won't get to the millions of threads for the various use-cases where we utilize `Debounce`, but it would be nice if `Debounce` consumed a constant amount of memory regardless how many actions were consumed in the debounce.

But these are potential improvements we'll consider once we actually have a problem to address, I think. Just food for thought for now.

tricorder/src/Tricorder/Watcher.hs
```diff
@@ -31,12 +33,28 @@ component =
...
-            pure [forever $ watchDirs dirs \path -> markDirty (changeKindFor path)]
+            let watches = sourceWatches dirs <> cabalWatches projectRoot
+            pure
+                [ watchFilePathsDebounced watches (markDirty . changeKindFor)
+                ]
         }
 
 
+sourceWatches :: [FilePath] -> [Watch]
+sourceWatches dirs = map (\d -> dirExt d ".hs" `excluding` containing "dist-newstyle") dirs
```
Gipphe, line 44: suggestion (non-blocking): Make excluded directories configurable.

If someone uses `stack` for example, they would like the `.stack-work` directory to also be excluded.

We could potentially also do something a bit more dynamic: exclude `.stack-work` only if we find a `stack.yaml` file in the repo.

But in general, allowing the user to supply additional exclusions if a good idea.

tricorder/test/Unit/Tricorder/FileWatcherSpec.hs
Gipphe, line 1: question: Can this file be merged into `test/Unit/Atelier/Effects/FileWatcherSpec.hs`?

Gipphe (approved):
LGTM, just some thoughts and a question.

atelier/src/Atelier/Effects/Debounce.hs
cgeorgii: We simply need a way to tell whether a new event's come in, so the index here is indeed beside the point.

atelier/src/Atelier/Effects/Debounce.hs
cgeorgii: Well, having a million events with the same key being fired in rapid succession seems wrong on other levels too.

I agree to defer this.

tricorder/src/Tricorder/Watcher.hs
cgeorgii: Good point. I've created [an issue](https://github.com/atelier-hub/tricorder/issues/26) to track this.

tricorder/test/Unit/Tricorder/FileWatcherSpec.hs
cgeorgii: Argh, yes, I forgot. Done.

PR #51, opened 2026-04-22
fix: daemon does not exit

Gipphe:


tricorder/src/Tricorder/Program.hs
```diff
@@ -345,21 +353,20 @@ showSource
...
     result <- querySource sockPath moduleNames
     case result of
         Left err -> Console.putTextLn $ "Error: " <> err
         Right results -> renderSourceResults results
 
 
 -- | Poll until the daemon socket becomes connectable.
-waitForSocket :: (Delay :> es, UnixSocket :> es) => FilePath -> Eff es ()
-waitForSocket sockPath = do
+waitForDaemon :: (Daemons :> es, Delay :> es, Reader PidFile :> es, UnixSocket :> es) => Eff es ()
```
cgeorgii: Hm, this `UnixSocket` effect is no longer needed. I don't get why the linter's not picking it up, though 🤔

PR #53, opened 2026-04-23
refactor: tag Daemons with a type-level tag, remove Reader dependency

cgeorgii:


Gipphe (commented):
This only really moves the responsibility of supplying the `PidFile` to the caller of the interpreter though. It'd be simpler to just make `runDaemon` have a `Reader PidFile`, and take it from there. Until the need arises where we actually have more than 1 daemon, I don't see how this improves things for us much at all, apart from introducing the possibility of typoes when specifying `Daemon tag :> es` constraints (which would result in missing interpreters, so it's not a runtime issue).

Also, I am working on completely revamping this to where we don't use `killAndWait` to kill the daemon, since it turns out that `SIGQUIT` (which is sent by `killAndWait`) does not always cause the daemon to exit, and `SIGTERM` does not always cause bracketed actions to clean up after themselves. Since the only way to ensure everything is cleaned up on the daemon's way out is to make it shut down of its own accord, we should instead send it a message to request it to shut down.

I'll detail all of this further in the PR I'm working on.

cgeorgii:
@Gipphe The issue with the current approach is that effectful might fail silently if there are two daemon interpreters in the stack. It will pick the first one in the stack, which might be incorrect and fail at runtime. The tag makes that impossible because we need to specify which daemon we mean at compilation time.

Gipphe (approved):
Truth be told, our `Daemon` effect should probably only really care about 1 potential daemon, and should instead pick the `PidFile` with `Reader PidFile` in its interpreter, thus preventing having to pass `PidFile` to each operation as we're doing now. But that's a small, insignificant change to make down the line should we feel the need to simplify further than this, so :shrug: 

I would like to see some way to use tags to pick out a particular `Daemon tag` effect and tie it to a corresponding hostname and port, but that's better done after [my PR](https://github.com/atelier-hub/tricorder/pull/57) probably.

PR #62, opened 2026-04-30
fix: use separate builddir for cabal repl to prevent spurious rebuilds

cgeorgii:
`cabal repl` and `cabal run` share `dist-newstyle/` by default. When the daemon starts its GHCi session, `cabal repl` overwrites the plan cache (`improved-plan`, `up-to-date`), causing `cabal run` to think a rebuild is needed even when no sources changed.

- Add `replBuildDir` field to `Config` (default `dist-newstyle/tricorder`)
- Pass `--builddir <replBuildDir>` in the auto-detected `cabal repl` command
- Configurable via `repl_build_dir:` in `.tricorder.yaml`

To verify, run this a few times and note how the program is not rebuilt:

`cabal run tricorder-exe -- stop; cabal run tricorder-exe -- start; cabal run tricorder-exe -- status --wait`

### DIFF
--- a/tricorder/src/Tricorder/Config.hs
+++ b/tricorder/src/Tricorder/Config.hs
@@ -55,6 +55,7 @@ data Config = Config
     , watchDirs :: [FilePath]
     , testTargets :: Maybe [Text]
     , outputFile :: Maybe FilePath
+    , replBuildDir :: FilePath
     }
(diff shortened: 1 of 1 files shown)


Gipphe (approved):
Love it 🚀

PR #75, opened 2026-05-07
feat: add dedicated daemon exe

Gipphe:
Depends on https://github.com/atelier-hub/tricorder/pull/78

I have a feeling that it might not be wise to have our effect stack carry through the `forkProcess` boundary that is `runDetached` in `Atelier.Effects.Posix.Daemons`. If we have any effect interpreters that spawn their own worker threads, these threads will _not_ be running in the daemon, for example.

As such, it is safer to have the daemon start up its own effect stack by having a dedicated `main`-like function for it to start from. This is somewhat in line with how we started the daemon in the past, but now it resides as a separate `Tricorder.Daemon.Main` module instead.

Since the daemon got this treatment, it seemed sensible to treat the CLI the same way, and as such `tricorder/app/Main.hs` was effectively copied to `tricorder/src/Tricorder/CLI/Main.hs`.

So with this PR, the `exe`s themselves just call their corresponding `Tricorder.(Daemon/CLI).Main.main` function.


tricorder/src/Tricorder/Daemon/Main.hs
```diff
@@ -0,0 +1,78 @@
...
+        . runDelay
+        . runFile
+        . runFileSystemIO
+        . runProjectRoot
+        . runRuntimeDir
+        . runSocketPath
+        . runLoadedConfig
+        . runSession
+        . runConfig @"observability" @Observability.Config
+        . runConfig @"observability.tracing" @TracingConfig
```
Gipphe, line 59: This will always return `TracingConfig`'s default value, since it doesn't inspect the nested property `"tracing"`, but instead looks for a top-level property named `"observability.tracing"`, of which there would be none.

This is fixed in https://github.com/atelier-hub/tricorder/pull/81.

cgeorgii (approved):

PR #76, opened 2026-05-07
feat: add `test-results` command

cgeorgii:
## Summary

Adds a new `tricorder test-results` command for inspecting the output of the latest test run without going through the full `status` output.

### New command

```
tricorder test-results [--failed] [--wait]
```

- **Default**: shows the full captured output for every test suite in the latest build, with GHCi/cabal startup noise stripped.
- **`--failed`**: shows only failing test suites; within each suite, shows individual failing test cases with their descriptions and failure details (parsed from hspec output). Falls back to full suite output with a warning for unrecognised runner formats.
- **`--wait`**: blocks until the current build and test cycle completes before printing results, matching the behaviour of `status --wait`.
- **Exit code**: 1 when any shown test suite failed or errored, 0 otherwise.

### Example output — `--failed`

```
test:tricorder-test  failed
  includes the one-liner prefix for an error:
    "E Foo.hs:10 type mismatch" does not contain "E Foo.hs:10 deliberate failure"
    Use -p '/includes the one-liner prefix for an error/' to rerun this test only.
```

When all suites passed (common use with `--wait --failed`):

```
All passed.
  test:my-tests
  test:other-tests
```

## Implementation notes

- **`Tricorder.TestOutput`** — new module with `parseHspecOutput`, which recognises right-aligned `OK`/`FAIL` lines in hspec output and collects indented failure detail lines. Returns `[]` for unrecognised formats (the `--failed` path detects this and falls back gracefully).
- **`BuildState.TestRun`** — extended with `testCases :: [TestCase]`, populated at capture time in `TestRunner`. `TestCase` and `TestCaseOutcome` are fully serialised (JSON), so the structured data is available everywhere downstream: `status --json`, the UI, and future commands.
- **GHCi noise stripping** — the full output path strips the cabal build profile, `[N of M] Compiling` lines, and the GHCi prompt/exit lines before display.
- **Skill doc** — `SKILL.md` updated with the new command, flag descriptions, example workflow step, and an explicit note not to run `cabal test` manually.

### DIFF
--- a/tricorder/src/Tricorder/CLI.hs
+++ b/tricorder/src/Tricorder/CLI.hs
@@ -2,6 +2,7 @@ module Tricorder.CLI
     ( showLog
     , showSource
     , showStatus
+    , showTests
     ) where
 
 import Data.Aeson (encode)
@@ -21,6 +22,7 @@ import Tricorder.Arguments
     ( FollowMode (..)
     , OutputFormat (..)
     , StatusOptions (..)
+    , TestOptions (..)
     , Verbosity (..)
     , WaitMode (..)
     )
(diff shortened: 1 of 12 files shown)


tricorder/src/Tricorder/Effects/TestRunner.hs
```diff
@@ -54,10 +55,10 @@ runTestRunnerIO act = do
                 pure (out, err)
             pure $ case result of
                 Left ex ->
-                    TestRun {target, outcome = TestsError (show ex :: Text), output = ""}
+                    TestRun {target, outcome = TestsError (show ex :: Text), output = "", testCases = []}
```
Gipphe: suggestion (non-blocking): Seems to me like `TestRun` should be an ADT aking to something like this:

```hs
data TestRun
  = TestsRunning {target :: Text}
  | TestsError {target :: Text, err :: SomeEception {- or Text -} }
  | TestsDone {target :: Text, testCases :: [TestCase]}
```

As it stands, the `TestRun` record includes a lot of fields that are not applicable in certain combinations. Example:

```hs
passedWithoutAnyCases = TestRun
  { target = "foo"
  , outcome = TestsPassed
  , testCases = []
  }

failedWithNoFailedCases = TestRun
  { target = "foo"
  , outcome = TestsFailed
  , testCases = [TestCase { description = "bar", outcome = TestCasePassed }]
  }
```

Is there a way we can make nonsense values of `TestRun` like this impossible?

tricorder/src/Tricorder/CLI.hs
```diff
@@ -159,6 +164,86 @@ showLog mLogFile followMode = case mLogFile of
...
+                    r.testRuns
+        in  if null r.testRuns then
+                Console.putStrLn "No test results."
+            else
+                if null runs then do
+                    Console.putStrLn "All passed."
+                    mapM_ (Console.putTextLn . ("  " <>) . (.target)) r.testRuns
+                else do
+                    mapM_ printTestOutput runs
+                    when (any (isFailed . (.outcome)) runs) exitFailure
```
Gipphe: nitpick: Just pass `r.testRuns` to this function, and it can be prettied up rather nicely.

```suggestion
    renderTestRuns [] = Console.putStrLn "No test results."
    renderTestRuns testRuns
        | null runs = do
              Console.putStrLn "All passed."
              for_ (Console.putTextLn . ("  " <>) . (.target)) testRuns
        | otherwise = do
              mapM_ printTestOutput runs
              when (any (isFailed . (.outcome)) runs) exitFailure
      where 
        runs = 
            if opts.failedOnly then
                filter (isFailed . (.outcome)) testRuns
            else
                testRuns
```

tricorder/src/Tricorder/Effects/TestRunner.hs
cgeorgii: Done in 00b71f494a70b36f5843a38c1aa2bef9cb218996. :ok_hand:

tricorder/src/Tricorder/CLI.hs
cgeorgii: Thanks, good suggestion.

PR #78, opened 2026-05-07
refactor: rename Config to Session

Gipphe:
It's been a bit confusing to me, until I started working to separate the effect stacks of the daemon and the CLI, that the `Config` record is used so sparingly. Seems that is because it is _only_ used by the daemon, and contains configuration that pertains to a given session/repo. As such, I suggest we rename this record to `Session` (or any better suggestion you might have) to better distinguish it from "configuration for everything" (which is currently in `Tricorder.Runtime`.)


cgeorgii (approved):
👍

PR #80, opened 2026-05-07
refactor: rework GhciSession component into Builder component

Gipphe:
Split GhciSession's `sessionListener` into multiple functions each with more concentrated responsibilities.

The new `buildWithGhciOnChange` will forcibly restart GHCi even if GHCi is in the middle of a reload.

Depends on https://github.com/atelier-hub/tricorder/pull/79


tricorder/src/Tricorder/Builder.hs
```diff
@@ -0,0 +1,373 @@
...
+       , State (Map FilePath [Diagnostic]) :> es
+       , State BuildId :> es
+       , Sub CabalChangeDetected :> es
+       , Sub SourceChangeDetected :> es
+       )
+    => MVar () -> Eff es Void
+buildWithGhciOnChange ref = forever do
+    put Map.empty
+    session <- ask @Session
+    projectRoot <- ask
```
cgeorgii, line 133: question: is there a reason to not include `projectRoot` in `Session`?

tricorder/src/Tricorder/Watcher.hs
```diff
@@ -59,9 +63,3 @@ isCabalFile :: FilePath -> Bool
 isCabalFile f =
     takeExtension f == ".cabal"
         || takeFileName f `elem` ["cabal.project", "package.yaml"]
-
-
-changeKindFor :: FilePath -> ChangeKind
-changeKindFor path
-    | isCabalFile path = CabalChange
-    | otherwise = SourceChange
```
cgeorgii, line 67: I think this was better placed here than in the `Builder`, don't you think?

tricorder/src/Tricorder/Builder.hs
```diff
@@ -0,0 +1,373 @@
...
+-- Run all configured test suites if the build has no errors.
+-- Transitions to 'Testing' phase while suites are running.
+runTestsIfClean
+    :: (Log :> es, Pub EnteringNewPhase :> es, TestRunner :> es)
+    => Session -> BuildId -> BuildResult -> Eff es [TestRun]
+runTestsIfClean (Session {testTargets}) bid partialResult
+    | null testTargets || any (\d -> d.severity == SError) partialResult.diagnostics = pure []
+    | otherwise = do
+        publish
+            $ EnteringNewPhase bid
```
cgeorgii, line 289: This looks like it would enable users to eventually extend tricorder with hooks. 👍

cgeorgii (commented):
Big PR. I'm going through it but my initial impression is that `Builder` includes some concerns that we should consider extracting (classifying types of files changed, handling diagnostics, etc.)

I'm okay with doing it in a subsequent PR, but it's something to keep in mind.

tricorder/src/Tricorder/Builder.hs
Gipphe: `Session` is daemon-specific stuff, and `projectRoot` is also used by the CLI. So we can either have the `ProjectRoot` separate from `Session`, ensuring the CLI and daemon get he project root dir the same way, or we can include it in `Session` to save a `Reader ProjectRoot` constraint. I opted for the former, but I don't have a particularly strong opinion either way.

tricorder/src/Tricorder/Watcher.hs
Gipphe: Considering it's only used in `Builder` in this particular PR, I think it's fine to have there.

There is an argument to be made about where change categorization should live. By moving that stuff into `Builder`, I think `Watcher` is very barebones, and `Builder` might have more responsibilities than we want, so I'm leaning towards moving `changeKindFor` and the corresponding stuff like `markWatchedFiles` back into `Watcher`.

tricorder/src/Tricorder/Builder.hs
Gipphe: That wasn't what I initially thought of when adding these events, but that is a very good use-case for these `Entering-` and `Entered-` forms of events.

cgeorgii (approved):

PR #84, opened 2026-05-11
feat: add version checking between client and daemon

cgeorgii:
- Embed git short hash at compile time via Template Haskell (`Tricorder.Version.gitHash :: Text`); `-fforce-recomp` ensures it stays fresh after every commit
- `--version` flag prints the hash and exits
- Every client query is wrapped in `ClientMessage { clientVersion, payload }`; the daemon logs a `WARN` on mismatch but continues serving
- Daemon logs its version on startup
- Version shown in the expanded info panel of the TUI (`g` to toggle)


tricorder/src/Tricorder/Effects/VersionCheck.hs
```diff
@@ -0,0 +1,39 @@
...
+
+
+runVersionCheck :: Eff (VersionCheck : es) a -> Eff es a
+runVersionCheck = interpret $ \_ -> \case
+    CheckVersion clientVersion ->
+        pure
+            $ if clientVersion == gitHash then
+                Right ()
+            else
+                Left VersionMismatch {expected = gitHash, received = clientVersion}
```
Gipphe: question: Since this interpreter is just a pure wrapper around a simple `checkVersion` function which requires _no_ additional effects, does it need to exist as a dedicated effect?

Could a `checkVersion :: Text -> Either VersionMismatch ()` just live in `Tricorder.Version`?

tricorder/src/Tricorder/UI/View.hs
```diff
@@ -111,12 +112,22 @@ viewHint = subtle $ txt "Press 'h' for help"
...
+        , viewLogFile di.logFile
+        , viewMetrics di.metricsPort
+        ]
+
+
+viewVersion :: Widget n
+viewVersion =
+    hBoxSpaced
+        1
+        [ emphasis $ txt "Version:"
```
Gipphe: suggestion: Make it clear that this is the CLI client version, not the daemon version, especially since this is in the "daemon info" section (which should probably be renamed to something more generic like "extended info" or something, I guess.)
```suggestion
        [ emphasis $ txt "Client version:"
```

Gipphe: todo (non-blocking): Might also be worthwhile to have the client ask the daemon what version it's running and display that in the ~~extended~~ daemon info to inform the user that there is a version mismatch. Food for a future PR.

tricorder/src/Tricorder/Version.hs
```diff
@@ -0,0 +1,27 @@
...
+import Language.Haskell.TH (litE, runIO, stringL)
+import System.Process (readProcess)
+
+
+-- | Short git hash of the commit this binary was built from.
+gitHash :: Text
+gitHash =
+    toText
+        ( $( do
+                hash <- runIO $ readProcess "git" ["rev-parse", "--short", "HEAD"] ""
```
Gipphe: thought: this is a problem I will have to contend with, but straight up using `git` here won't work for me in a jujutsu workspace, since jujutsu does not yet colocate `.git` in workspaces. This is solved by using the `GIT_DIR` env var though, so it might not pose a problem after all.

tricorder/src/Tricorder/Effects/VersionCheck.hs
cgeorgii: Ah, that is a good point. I changed the initial implementation completely and now it's simple enough that it absolutely doesn't need it. Forgot to remove it.

tricorder/src/Tricorder/UI/View.hs
cgeorgii: Yes, but it requires changing the wire protocol so I decided to defer it.

tricorder/src/Tricorder/Version.hs
cgeorgii: Hm, good point. Nowadays more and more people are using different version control systems, so this might become a problem in the future. Not sure what a more robust solution would look like, though.

Gipphe: Pass it as a compilation option of some sorts? Isn't that how GHC does it, for example?

Gipphe: I misunderstood what you meant by env var. I thought you meant reading the env var at _runtime_, not passing it as an env var to be read in TH at compile time 😅  Makes more sense now.

Gipphe (approved):

PR #86, opened 2026-05-12
fix: pass resolved targets to resolveTestTargets

cgeorgii:
## Problem

When targets are not explicitly configured, tricorder auto-detects them from the `.cabal` file via `resolveTargets`. However, `resolveTestTargets` and `resolveWatchDirs` were both reading `cfgFile.targets` directly — the raw config value, which is empty before resolution. This meant:

- **Tests never ran**: `resolveTestTargets` filtered an empty list for `test:` components, always producing `[]`
- **Wrong watch directories**: `resolveWatchDirs` passed the empty list to `resolveWatchDirsFromTargets`, which fell back to `["."]` (the entire project root) instead of inferring source dirs from the cabal file

Both functions already had the right logic — they just received the wrong input.

## Changes

- `resolveTestTargets` now takes the resolved target list as an explicit argument instead of reading `cfg.targets`
- `resolveWatchDirs` does the same
- Both are called in `runSession` with `effectiveTargets` (the post-resolution list) rather than the raw config field
- Added `resolveWatchDirs` unit tests covering: explicit `watch_dirs` config, empty-targets fallback, source dir inference from resolved targets, and no-cabal-file fallback

### DIFF
--- a/tricorder/src/Tricorder/Session.hs
+++ b/tricorder/src/Tricorder/Session.hs
@@ -119,11 +119,11 @@ detectCommand targets replBuildDir (ProjectRoot projectRoot) = do
 -- 1. @watch_dirs@ from config, if non-empty (used as-is relative to project root)
 -- 2. @hs-source-dirs@ inferred from cabal targets, if targets are set
 -- 3. Falls back to @["."]@ (project root) if neither is available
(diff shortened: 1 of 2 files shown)


Gipphe (approved):

Gipphe (approved):
Good catches all around.

PR #89, opened 2026-05-12
feat: add configurable timeout for test suite runs

cgeorgii:
- Add `testTimeout :: Int` (seconds) to `Session.Config` and `Session`, defaulting to 10s; set `test-timeout: 0` in `.tricorder.yaml` to disable
- Replace `readProcess` in `runTestRunnerIO` with `withProcessTerm` + `withTimeout`, sending SIGTERM and returning `TestRunErrored` on expiry
- Add `withTimeout` to `Atelier.Effects.Delay` as a general utility that races an action against a `Delay`-based deadline, tested in `DelaySpec`

### DIFF
--- a/atelier/src/Atelier/Effects/Delay.hs
+++ b/atelier/src/Atelier/Effects/Delay.hs
@@ -2,6 +2,7 @@ module Atelier.Effects.Delay
     ( Delay
     , wait
     , every
+    , withTimeout
     , runDelay
     , runDelayNoOp
 
@@ -18,6 +19,7 @@ module Atelier.Effects.Delay
 import Data.Time.Units (TimeUnit, convertUnit, toMicroseconds)
 import Effectful (Effect, IOE)
 import Effectful.Concurrent (Concurrent, threadDelay)
+import Effectful.Concurrent.Async (race)
 import Effectful.Concurrent.MVar (newEmptyMVar, takeMVar)
 import Effectful.Dispatch.Dynamic (interpret_)
 import Effectful.State.Static.Shared (State, modifyM)
@@ -43,6 +45,16 @@ every :: (Delay :> es, TimeUnit t) => t -> Eff es () -> Eff es Void
 every delay action = forever do
     action
     wait delay
+
+
+-- | Race an action against a deadline. Returns 'Right' if the action
+-- completes first, 'Left' if the timeout fires first.
+withTimeout
+    :: (Concurrent :> es, Delay :> es, TimeUnit t)
+    => t
+    -> Eff es a
+    -> Eff es (Either () a)
+withTimeout duration = race (wait duration)
 
 
 runDelay :: (Concurrent :> es) => Eff (Delay : es) a -> Eff es a
--- a/tricorder/src/Tricorder/Effects/TestRunner.hs
+++ b/tricorder/src/Tricorder/Effects/TestRunner.hs
@@ -13,20 +13,25 @@ module Tricorder.Effects.TestRunner
     ) where
 
 import Control.Exception (throwIO)
+import Data.Time.Units (Second)
 import Effectful (Effect, IOE)
+import Effectful.Concurrent (Concurrent)
+import Effectful.Concurrent.STM (atomically)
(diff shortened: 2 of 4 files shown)


atelier/src/Atelier/Effects/Delay.hs
```diff
@@ -45,6 +47,16 @@ every delay action = forever do
...
 
 
+-- | Race an action against a deadline. Returns 'Right' if the action
+-- completes first, 'Left' if the timeout fires first.
+withTimeout
+    :: (Concurrent :> es, Delay :> es, TimeUnit t)
+    => t
+    -> Eff es a
+    -> Eff es (Either () a)
+withTimeout duration = race (wait duration)
```
Gipphe, line 57: suggestion (non-blocking): If we want to stick with `Conc` and `Ki`, we can implement a `race` function somewhat easily.

This is just something I have in the branch I'm working on:

```hs
race :: (Conc :> es) => Eff es a -> Eff es b -> Eff es (Either a b)
race m1 m2 = scoped do
    ref <- unsafeEff_ IO.newEmptyMVar
    _ <- fork do
        res <- await =<< fork m1
        unsafeEff_ $ IO.putMVar ref $ Left res
    _ <- fork do
        res <- await =<< fork m2
        unsafeEff_ $ IO.putMVar ref $ Right res
    unsafeEff_ $ IO.readMVar ref
```

An entirely `ki`-based `race` function, using sub scopes. I leave this suggestion as non-blocking because I don't actually see a good argument to use a version of `race` implemented with `ki`, so 🤷.

Gipphe: Also, what about `Effectful.Timeout`? That's probably a more fitting thing to use in this case.

tricorder/src/Tricorder/Effects/TestRunner.hs
```diff
@@ -42,23 +47,36 @@ makeEffect ''TestRunner
...
-            result <- trySync $ liftIO do
-                let config =
-                        setStdin (byteStringInput ":main\n:quit\n")
-                            $ setWorkingDir projectRoot
-                            $ proc "cabal" ["repl", toString target]
-                (_, out, err) <- readProcess config
-                pure (out, err)
+            Session {testTimeout} <- ask
+            let config =
+                    proc "cabal" ["repl", toString target]
```
Gipphe, line 57: todo (non-blocking): Since we support `stack` for building, we should probably also support using `stack test`.

Just something I noticed when reviewing this PR. Not something to deal with here.

tricorder/src/Tricorder/Effects/TestRunner.hs
```diff
@@ -42,23 +47,36 @@ makeEffect ''TestRunner
...
+            result <- trySync $ withProcessTerm config \p -> do
+                let collectOutput = do
+                        out <- atomically (getStdout p)
+                        err <- atomically (getStderr p)
+                        pure (out <> err)
+                case testTimeout of
+                    secs | secs <= 0 -> Right <$> collectOutput
+                    secs ->
+                        withTimeout (fromIntegral secs :: Second) collectOutput >>= \case
+                            Left () -> liftIO (stopProcess p) >> pure (Left secs)
```
Gipphe: nitpick (non-blocking): According to [withProcessTerm's docs](https://hackage-content.haskell.org/package/typed-process-0.2.13.0/docs/System-Process-Typed.html#v:withProcessTerm), it calls `stopProcess` automatically once the callback returns, so this `stopProcess` call might be redundant.

Gipphe (approved):

atelier/src/Atelier/Effects/Delay.hs
cgeorgii: I didn't go with `Effectful.Timeout` because it's not testable.

tricorder/src/Tricorder/Effects/TestRunner.hs
cgeorgii: Good catch, I should've suspected from the name :P

PR #98, opened 2026-05-15
feat: re-resolve targets on cabal change

### LINKED ISSUE #95: Cabal file changes can result in missed targets
When adding a new build target to a cabal file, the daemon will restart `GHCi`, but it will not evaluate `targets` and `testTargets` again, and the new target will _not_ be built by `tricorder`.

Gipphe:
- **Session re-resolution on cabal change**: introduces a SessionStore effect that owns the live Session value and
  reloads it when a .cabal file changes, replacing the static Reader Session that required a daemon restart to pick up
  new targets
- **withSubSession restart primitive**: generalises the reload loop — BuilderSession and WatcherSession are projections
  of Session that trigger restarts only when their relevant fields actually change, avoiding unnecessary GHCi
  interruptions
- **Reloader handle**: passed to managed components so they can trigger a session reload directly (e.g.
  restartOnCabalChange calls reloader.reload after detecting a cabal change)
- **SessionStoreReloaded event**: published after each reload; drives the restart logic in withBuilderSession and
  withWatcherSession
- **Utility additions**: Conc.race (return whichever of two computations finishes first) and Publishing.listenOnce /
  listenOnce_ (await a single event)

Closes https://github.com/atelier-hub/tricorder/issues/95
Depends on https://github.com/atelier-hub/tricorder/pull/96

### DIFF
--- a/atelier/src/Atelier/Effects/Conc.hs
+++ b/atelier/src/Atelier/Effects/Conc.hs
@@ -12,6 +12,7 @@ module Atelier.Effects.Conc
     , Scope (..)
     , scoped
     , restartableForkWith
+    , restartableForkLoop
 
       -- * Interpreters
     , runConcBase
@@ -75,6 +76,18 @@ restartableForkWith signal setup action = forever $ scoped do
     r <- setup
     _ <- fork (action r)
     signal
+
+
+-- | Like 'restartableForkWith', but threads a value across iterations: the
+-- signal returns the next @r@, which is passed to the next fork. The initial
+-- @r@ seeds the first fork.
+restartableForkLoop :: (Conc :> es) => r -> Eff es r -> (r -> Eff es a) -> Eff es Void
+restartableForkLoop initial signal action = go initial
+  where
+    go r = scoped do
+        _ <- fork (action r)
+        r' <- signal
+        go r'
 
 
 -- | Base interpreter: resolves 'Conc' operations using Ki.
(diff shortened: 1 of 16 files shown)


atelier/src/Atelier/Effects/Conc.hs
```diff
@@ -50,6 +53,20 @@ newtype Thread a = Thread (Ki.Thread a)
 makeEffect ''Conc
 
 
+-- | Forks the two computations and returns the value of the computation that
+-- completed first.
+race :: (Conc :> es) => Eff es a -> Eff es b -> Eff es (Either a b)
+race a b = scoped do
+    ref <- unsafeEff_ newEmptyMVar
```
cgeorgii: 🤔

Gipphe: Added a comment rationalizing the use of `unsafeEff_` here.

Since the `MVar` in this function lives _only_ for the duration of this function, we probably want to hide it from callers of `race` so they don't have to keep `Concurrent` in their stack unnecessarily. Since `Conc` can only be resolved through `runConcBase` and `runConc`, you are forced to resolve `IOE` regardless.

In short: `Conc` is a more side-effecty effect than using an `MVar` locally, so we just hide the `MVar` in `race` as an implementation detail.

cgeorgii (approved):
Nice :)

PR #100, opened 2026-05-19
refactor: add custom GHCi backend and isolate ghcid dependencies

cgeorgii:
- Add `Tricorder.Internal.GhciParser`: standalone parser for raw GHCi output (`:reload`, `:show modules`) into structured `GhciLoad` items, with a `LoadResult` type and `toRelative`/`extractTitle` utilities
- Add `Tricorder.Internal.GhciProcess`: subprocess-based GHCi session manager using `system-process-typed`, with a marker-based sync protocol to drive `:reload` / `:show modules` without the ghcid library
- Add `Tricorder.Internal.GhcidBackend`: isolates the four ghcid-library helpers (`collectResult`, `stopGhciSilently`, `toDiagnostics`, `parseProgress`) so the old backend can be deleted as a unit later
- Add `runGhciSessionCustom` interpreter in `GhciSession`, selectable via `TRICORDER_GHCI_BACKEND=custom`; `runGhciSessionFromEnv` picks between the two backends at runtime
- Add `GhciParserSpec` unit tests covering loading lines, diagnostic spans, span formats, ANSI stripping, and `:show modules` parsing
- Add proposal 005 docs (research, design, spec)


tricorder/proposals/005-replace-ghcid/design.md
```diff
@@ -0,0 +1,189 @@
...
+### Configuration
+
+`GhciProcess.Config` holds `startupTimeout`, `shutdownTimeout`, and
+`extraSetupCommands`. `withGhciProcess` is called with `def` (the `Default` instance)
+in `runGhciSessionCustom`; callers that need non-default timeouts can call
+`withGhciProcess` directly.
+
+### Command serialisation
+
+An `MVar ()` (acting as a mutex) serialises access to the subprocess stdin. Only one
```
Gipphe: question: Should this be a `Semaphore` instead?

tricorder/src/Tricorder/Effects/GhciSession.hs
```diff
@@ -160,145 +145,32 @@ runGhciSessionScripted results = reinterpret (evalState results) $ \env ->
...
+runGhciSessionCustom :: (BuildStore :> es, Concurrent :> es, Delay :> es, File :> es, IOE :> es) => Eff (GhciSession : es) a -> Eff es a
+runGhciSessionCustom = interpret $ \env -> \case
+    WithGhci cmd (ProjectRoot dir) handler ->
+        withGhciProcess def cmd dir \proc ->
+            localLift env (ConcUnlift Persistent Unlimited) \liftEff ->
+                localUnlift env (ConcUnlift Persistent Unlimited) \unlift -> do
+                    let doReload = liftEff
+                            $ reloadGhci proc dir \l -> do
+                                bs <- getState
+                                setPhase bs.buildId
```
Gipphe: suggestion: I think it'd be wise to reduce the scope of responsibilities for `GhciSession` and let the caller set the phase in the `BuildStore`.

This module shouldn't need to concern itself with the `BuildStore` at all.

tricorder/src/Tricorder/Effects/GhciSession.hs
```diff
@@ -160,145 +145,32 @@ runGhciSessionScripted results = reinterpret (evalState results) $ \env ->
...
+                                    $ Building
+                                    $ Just
+                                    $ BuildProgress {compiled = l.index, total = l.total}
+                    initialResult <- unlift doReload
```
Gipphe: question (non-blocking): Doesn't GHCi build initially to begin with? Performing a reload here should be redundant.

Not that it would matter much for performance since reloading immediately after a build is practically a no-op.

tricorder/src/Tricorder/Internal/GhciParser.hs
```diff
@@ -0,0 +1,365 @@
+module Tricorder.Internal.GhciParser
```
Gipphe: question: Would these modules in `Tricorder.Internal` be better placed in `Tricorder.Effects.GhciSession`? In a `GhciSession` directory, that is?

These modules pertain to the `GhciSession` module very directly, and we don't have a convention of naming module paths `Internal`. This library is effectively an "application library", and not directly built to be used as an actual library, so we don't really have a concept of "internal" modules with a non-stable API.

tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
Gipphe, line 1: comment: I will admit that reviewing a parser for some output you don't really know the structure of is a futile effort, so I'll trust this parser is sensible for GHCi's output 👍

tricorder/src/Tricorder/Internal/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+                , stateVar = stateVar
+                }
+
+    -- Sync: drain until marker 1 seen, then set counter to 2
+    let marker1 = markerFor 1
+    sendSyncCommand inp marker1
+    _ <-
+        concurrently
+            (drainUntil out marker1)
+            (drainUntil err marker1)
```
Gipphe: suggestion (non-blocking): Use `Conc`.

```suggestion
    _ <- Conc.scoped do
             void $ Conc.fork $ drainUntil out marker1
             void $ Conc.fork $ drainUntil err marker1
             Conc.awaitAll
```

Only good reason why we should use `Conc` here is to tear down these two threads if `startGhciProcess` is thrown an exception while they're running, and we want these threads to halt _immediately_ instead of draining the GHCi output first. If something threw an exception at `startGhciProcess` it's likely that these two threads would experience issues of their own seeing as the GHCi process would probably have been stopped by `withGhciProcess`' `bracket` though.

tricorder/src/Tricorder/Internal/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+            liftIO $ TIO.hPutStrLn ghciProcess.stdin command
+            File.hFlush ghciProcess.stdin
+            sendSyncCommand ghciProcess.stdin marker
+            (stdoutLines, stderrLines) <-
+                concurrently
+                    (drainUntil ghciProcess.stdout marker)
+                    (drainUntil ghciProcess.stderr marker)
+            pure (stdoutLines ++ stderrLines)
+        )
+        `finally` atomically (writeTVar ghciProcess.stateVar (Idle (n + 1)))
```
Gipphe: nitpick:

```suggestion
    doExec n `finally` atomically (writeTVar ghciProcess.stateVar (Idle (n + 1)))
  where
    doExec n = do
        let marker = markerFor n
        liftIO $ TIO.hPutStrLn ghciProcess.stdin command
        File.hFlush ghciProcess.stdin
        sendSyncCommand ghciProcess.stdin marker
        (stdoutLines, stderrLines) <-
            concurrently
                (drainUntil ghciProcess.stdout marker)
                (drainUntil ghciProcess.stderr marker)
        pure (stdoutLines ++ stderrLines)
```

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+                    (drainUntil ghciProcess.stdout marker)
+                    (drainUntil ghciProcess.stderr marker)
+            pure (stdoutLines ++ stderrLines)
+        )
+        `finally` atomically (writeTVar ghciProcess.stateVar (Idle (n + 1)))
+
+
+-- | Interrupt the currently running GHCi command (if any).
+--
+-- Sends SIGINT to the GHCi process group and writes a new sync marker so
```
Gipphe, line 216: question: Since we're sending a `SIGINT`, which I understand to be equivalent to `Ctrl + c`, GHCi won't exit if it's currently idle, right?

tricorder/src/Tricorder/Internal/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+    sendSyncCommand ghciProcess.stdin (markerFor (n + 1))
+
+
+-- | Stop the GHCi process gracefully, falling back to forced termination.
+--
+-- Never throws — all errors are swallowed.
+stopGhciProcess :: (Concurrent :> es, Delay :> es, File :> es, IOE :> es) => Config -> GhciProcess -> Eff es ()
+stopGhciProcess config ghciProcess = do
+    -- Try to write :quit
+    void $ try @SomeException $ do
```
Gipphe: suggestion (blocking): Use `Effectful.Exception.trySync` to not catch async exceptions.

Unless that is explicitly what you want to do here, but I don't quite see a reason to do that here.

```suggestion
    void $ trySync do
```

tricorder/src/Tricorder/Internal/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+
+    -- Wait up to shutdownTimeout seconds for the process to exit, then force-kill
+    result <- withTimeout config.shutdownTimeout (liftIO $ waitExitCode ghciProcess.handle)
+    when (not (isRight result))
+        $ liftIO
+        $ stopProcess ghciProcess.handle
+
+    -- Close all handles, ignoring errors
+    for_ [ghciProcess.stdin, ghciProcess.stdout, ghciProcess.stderr] \h ->
+        void $ try @SomeException $ File.hClose h
```
Gipphe: suggestion (blocking): Ditto as above: use `Effectful.Exception.trySync` to prevent accidentally catching async exceptions as `SomeException`.

```suggestion
        void $ trySync $ File.hClose h
```

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+    -- Close all handles, ignoring errors
+    for_ [ghciProcess.stdin, ghciProcess.stdout, ghciProcess.stderr] \h ->
+        void $ try @SomeException $ File.hClose h
+
+
+-- ---------------------------------------------------------------------------
+-- Internal helpers
+
+-- | Build the finish marker for counter value @n@.
+markerFor :: Int -> Text
```
Gipphe, line 254: nitpick (non-blocking): Should this take a `Word` instead of an `Int`?

This is a bit of a debate I'm having with myself quite often: a lot of the time, I have a function that only makes sense for positive integers. So `Word` makes sense to use, but carries the extra burden of having to `fromInteger` it back and forth whenever I sit there with an `Int`. What is your take on this? I like to keep functions as total as possible. `markerFor` does technically work for negative integers, producing markers such as `$~TRI-FINISH--1~#` for `-1`, so I guess since this is effectively just a piece of text we're matching on, `markerFor` can take an `Int`.

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+--
+-- Stops on any line containing the marker prefix, so an interrupt that writes
+-- a later marker will unblock a drain waiting for an earlier one.
+-- Returns accumulated non-marker lines in order. Throws 'UnexpectedExit' on
+-- EOF before a marker is seen.
+drainUntil :: (File :> es) => Handle -> Text -> Eff es [Text]
+drainUntil h command = go []
+  where
+    go acc = do
+        result <- try @E.IOException $ File.hGetLine h
```
Gipphe, line 290: note: Contrary to what I've said previously about `try @SomeException`, this is a more correct use of Effectful's `try`, where you specifically catch an exception type that is synchronous.

Just felt I had to comment that, since I feel like I'm nagging a lot about `try @SomeException` 😅

tricorder/src/Tricorder/Internal/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+-- Returns accumulated non-marker lines in order. Throws 'UnexpectedExit' on
+-- EOF before a marker is seen.
+drainUntil :: (File :> es) => Handle -> Text -> Eff es [Text]
+drainUntil h command = go []
+  where
+    go acc = do
+        result <- try @E.IOException $ File.hGetLine h
+        case result of
+            Left ex ->
+                if isEOFError ex then
```
Gipphe: suggestion: Drop this conditional unless we want to do something different on `EOF`.

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+        case result of
+            Left ex ->
+                if isEOFError ex then
+                    throwIO $ UnexpectedExit command (listToMaybe (reverse acc))
+                else
+                    throwIO $ UnexpectedExit command (listToMaybe (reverse acc))
+            Right line ->
+                if markerPrefix `T.isInfixOf` line then
+                    pure (reverse acc)
```
Gipphe, line 296: suggestion (non-blocking): Consider using something like [Seq](https://hackage-content.haskell.org/package/containers-0.8/docs/Data-Sequence.html#t:Seq).

`Seq` has `O(1)` complexity on appending values, so it might be a better fit here compared to prepending then reversing the list of lines. Wholly depends on whether this is actually, genuinely a performance consideration for us, which might only be the case for large codebases with many diagnostic messages at once.

tricorder/src/Tricorder/Internal/GhciProcess.hs
```diff
@@ -0,0 +1,335 @@
...
+    isVersionLine line =
+        let stripped = stripAnsi line
+        in  "GHCi, version " `T.isInfixOf` stripped
+                || "GHCJSi, version " `T.isInfixOf` stripped
+                || "Clashi, version " `T.isInfixOf` stripped
+
+    go = do
+        result <- try @E.IOException $ File.hGetLine h
+        case result of
+            Left _ -> pure ()
```
Gipphe: question: This will cause exceptions raised by `hGetLine` to be interpreted as if the banner was seen. Is that the desired result?

tricorder/test/Unit/Tricorder/Internal/GhciParserSpec.hs
```diff
@@ -0,0 +1,268 @@
...
+                       ]
+
+    it "handles padded module index (e.g. [ 1 of 47])" do
+        let input =
+                [ "[ 1 of 47] Compiling Main              ( app/Main.hs, interpreted )"
+                , "Ok, 1 module loaded."
+                ]
+        parseReload input `shouldBe` [GLoading GhciLoading {index = 1, total = 47, moduleName = "Main", sourceFile = "app/Main.hs"}]
+
+    it "returns empty list when only summary line" do
```
Gipphe: nitpick:

```suggestion
    describe "when only summary line" $ it "returns empty list" do
```

tricorder/test/Unit/Tricorder/Internal/GhciParserSpec.hs
```diff
@@ -0,0 +1,268 @@
...
+                       ]
+
+
+--------------------------------------------------------------------------------
+-- parseReload: -fhide-source-paths output
+--------------------------------------------------------------------------------
+
+testHideSourcePaths :: Spec
+testHideSourcePaths = do
+    it "produces no GLoading items when source paths are hidden" do
```
Gipphe: nitpick:

```suggestion
    describe "when source paths are hidden" $ it "produces no GLoading items" do
```

tricorder/test/Unit/Tricorder/Internal/GhciParserSpec.hs
```diff
@@ -0,0 +1,268 @@
...
+                                , startPos = (10, 5)
+                                , endPos = (10, 5)
+                                , messageLines =
+                                    [ "src/Foo.hs:10:5: error: Variable not in scope: foo"
+                                    , "    Perhaps you meant: 'bar'"
+                                    ]
+                                }
+                       ]
+
+    it "returns empty list when all modules are already up to date" do
```
Gipphe: nitpick:

```suggestion
    describe "when all modules are already up to date" $ it "returns empty list" do
```

package.yaml
```diff
@@ -86,6 +86,7 @@ library:
   - filepath
   - ghcid
   - hashable
+  - async
```
Gipphe: question: Isn't this dependency unused now?

tricorder/src/Tricorder/Effects/GhciSession.hs
cgeorgii: I agree, but this is already the case with the current implementation. Let's defer this refactor.

tricorder/src/Tricorder/Internal/GhciParser.hs
cgeorgii: Yes, you're right. I also don't like it, I'll move them.

tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
cgeorgii: I added tests as much as possible :shrug:

tricorder/src/Tricorder/Internal/GhciProcess.hs
cgeorgii: Good call. I'll apply the suggestion.

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
cgeorgii: Argh, I changed this one to `IOException` but I forgot the others. I'll finish the cleanup, thanks for catching it.

package.yaml
cgeorgii: Whoops, yes.

Gipphe (changes requested):
Leave your hands off of those poor, innocent async exceptions, and this PR is pretty much good to go, apart from some minor nits, suggestions and questions 👌

tricorder/proposals/005-replace-ghcid/design.md
cgeorgii: No, the design doc was stale. I switched to a `TVar SessionState` and forgot to update it.

tricorder/src/Tricorder/Effects/GhciSession.hs
cgeorgii: Yeah, the cost is basically non-existent so let's keep this for now and refactor later.

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
Gipphe: Deferred until we find that we need it.

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
```diff
@@ -0,0 +1,333 @@
...
+  where
+    doExec n = do
+        let marker = markerFor n
+        liftIO $ TIO.hPutStrLn ghciProcess.stdin command
+        File.hFlush ghciProcess.stdin
+        sendSyncCommand ghciProcess.stdin marker
+        (stdoutLines, stderrLines) <-
+            concurrently
+                (drainUntil ghciProcess.stdout marker)
+                (drainUntil ghciProcess.stderr marker)
```
Gipphe, line 210: suggestion (non-blocking): Forgot about this use of `concurrently` when I suggested `Conc` earlier.

```suggestion
        (stdoutLines, stderrLines) <- Conc.scoped do
            stdout <- Conc.fork $ drainUntil ghciProcess.stdout marker
            stderr <- Conc.fork $ drainUntil ghciProcess.stderr marker
            (,) <$> Conc.await stdout <*> Conc.await stderr
```

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
Gipphe: Should the others be `try @IOException` as well, or should they be `trySync`?

Gipphe (approved):

PR #102, opened 2026-05-19
refactor: rewrite GhciParser using megaparsec

cgeorgii:
- Add `LineStream` newtype (`Token = Text`) for line-oriented parsing over `[Text]` without orphan instances
- Replace hand-rolled `Text` manipulation in `parseReload` and `parseShowModules` with megaparsec combinators; `span isMessageBody` becomes `many (satisfy isMessageBody)`
- Convert `parseLoadingLine` and `showModuleLine` to `TextParser` helpers run via `runTP`
- Add megaparsec to build-depends; update nix materialized files


tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
```diff
@@ -74,6 +86,277 @@ data LoadResult = LoadResult
...
+positionP = parenForm <|> simpleForm
+  where
+    parenForm = do
+        _ <- char '('
+        l1 <- L.decimal
+        _ <- char ','
+        c1 <- L.decimal
+        _ <- char ')'
+        _ <- char '-'
+        _ <- optional (char '(') -- some GHCi versions emit "(L1,C1)-((L2,C2):"
```
Gipphe, line 153: question: Is this missing an opening parens?

Since some versions emit `((L2,C2):`, we want to consume the first `(`, and then _optionally_ consume the second `(`.

```suggestion
        _ <- char '('
        _ <- optional (char '(') -- some GHCi versions emit "(L1,C1)-((L2,C2):"
```

tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
```diff
@@ -74,6 +86,277 @@ data LoadResult = LoadResult
...
+        c2 <- L.decimal
+        _ <- char ')'
+        _ <- char ':'
+        pure (Position l1 c1, Position l2 c2)
+    simpleForm = do
+        l <- L.decimal
+        _ <- char ':'
+        c <- L.decimal
+        choice
+            [ do _ <- char ':'; pure (Position l c, Position l c)
```
Gipphe: suggestion (non-blocking): Reformat this for readability.

either

```suggestion
            [ do
                _ <- char ':'
                pure (Position l c, Position l c)
```

or 

```suggestion
            [ char ':' *> pure (Position l c, Position l c)
```

tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
```diff
@@ -74,6 +86,277 @@ data LoadResult = LoadResult
...
+            , do
+                _ <- char '-'
+                c2 <- L.decimal
+                _ <- char ':'
+                pure (Position l c, Position l c2)
+            ]
+
+
+-- | Parse the full diagnostic header:
+--   @file:pos:@ or @drive:path:pos:@ for Windows.
```
Gipphe, line 177: comment: Interesting that we support Windows in this parser, when the rest of `tricorder` does not support Windows 😉

Gipphe (approved):
I find `megaparsec` parsers to generally be very readable, and this is no exception. LGTM 💯

tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
cgeorgii: Claude added it and I figured why not.

tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
cgeorgii: Added a test for this too.

PR #103, opened 2026-05-19
feat: add test results view to UI under T key

cgeorgii:
- add test results view to UI under T key
  - Replaces separate `daemonInfoView`/`showHelp` fields with a `[ActiveView]` stack so any view can be layered and `Esc` always pops one level
  - `t` cycles through fail-only → full → off by pushing/replacing/popping on the stack; `g` and `h` push/pop their views the same way
  - All layer dispatch is consolidated in `view`; adding a new view only requires a new `ActiveView` constructor and one new branch there
  - `TestView` uses `Bounded`/`Enum` deriving so cycling is free from the constructor order rather than explicit pattern matches
  - Suite summary lines show `passed (N)` or `X/N failed` when `testCases` is populated (hspec output); falls back to plain pass/fail otherwise
- exclude compilation time from test timeout
  - Adds `durationMs :: Maybe Int` to `TestRunCompletion`, populated by `parseHspecDuration`
  - Parses non-indented summary lines ending with `(Xs)`, matching both
    `All N tests passed (0.33s)` and `N out of M tests failed (0.06s)`
  - CLI `status` and `test-results` append the formatted duration to the pass/fail text
  - TUI `viewCompletionStatus` shows it as a dimmed suffix widget
  - Streams `cabal repl` stdout via `createPipe` instead of buffering
  - Prepends a sentinel expression (`putStrLn "#~TRI-TEST-READY~#"`) to
    stdin; `drainUntilReady` accumulates output until the sentinel appears,
    covering the full startup and compilation phase without a timeout
  - `testTimeout` now applies only to the remaining stdout read after the
    sentinel — i.e. test execution only
  - Switches to `TIO.hGetLine`/`TIO.hGetContents` (Text-based IO), removing
    the `BSL` encode/decode round-trip for the combined output
- exclude compilation time from test timeout
  - Streams `cabal repl` stdout via `createPipe` instead of buffering
  - Prepends a sentinel expression (`putStrLn "#~TRI-TEST-READY~#"`) to
    stdin; `drainUntilReady` accumulates output until the sentinel appears,
    covering the full startup and compilation phase without a timeout
  - `testTimeout` now applies only to the remaining stdout read after the
    sentinel — i.e. test execution only
  - Switches to `TIO.hGetLine`/`TIO.hGetContents` (Text-based IO), removing
    the `BSL` encode/decode round-trip for the combined output

### DIFF
--- a/atelier/src/Atelier/Time.hs
+++ b/atelier/src/Atelier/Time.hs
@@ -11,11 +11,12 @@ module Atelier.Time
 
       -- * Conversions
     , nominalDiffTime
+    , fromMicroseconds
     , toMicroseconds
     , convertUnit
     ) where
 
-import Data.Aeson (FromJSON (..))
+import Data.Aeson (FromJSON (..), ToJSON (..))
 import Data.Time (NominalDiffTime)
 import Data.Time.Units (Hour, Microsecond, Millisecond, Minute, Second, TimeUnit, convertUnit, fromMicroseconds, toMicroseconds)
 
@@ -45,3 +46,23 @@ instance FromJSON Minute where
 
 instance FromJSON Hour where
     parseJSON = fmap fromInteger . parseJSON
+
+
+instance ToJSON Microsecond where
+    toJSON us = toJSON (toMicroseconds us)
+
+
+instance ToJSON Millisecond where
+    toJSON ms = toJSON (toMicroseconds ms `div` 1000)
+
+
+instance ToJSON Second where
+    toJSON s = toJSON (toMicroseconds s `div` 1_000_000)
+
+
+instance ToJSON Minute where
+    toJSON m = toJSON (toMicroseconds m `div` 60_000_000)
+
+
+instance ToJSON Hour where
+    toJSON h = toJSON (toMicroseconds h `div` 3_600_000_000)
--- a/tricorder/src/Tricorder/CLI/Render.hs
+++ b/tricorder/src/Tricorder/CLI/Render.hs
@@ -8,6 +8,7 @@ module Tricorder.CLI.Render
     ) where
 
 import Atelier.Effects.Console (Console)
+import Atelier.Time (Millisecond, toMicroseconds)
 import Tricorder.BuildState
     ( Diagnostic (..)
     , Severity (..)
@@ -18,12 +19,13 @@ import Tricorder.SourceLookup (ModuleSourceResult (..))
 import Atelier.Effects.Console qualified as Console
 
 
-formatDuration :: Int -> Text
-formatDuration ms =
-    if ms < 1000 then
-        show ms <> "ms"
-    else
-        show (ms `div` 1000) <> "." <> show ((ms `mod` 1000) `div` 100) <> "s"
+formatDuration :: Millisecond -> Text
+formatDuration d =
+    let ms = toMicroseconds d `div` 1000
+    in  if ms < 1000 then
+            show ms <> "ms"
+        else
+            show (ms `div` 1000) <> "." <> show ((ms `mod` 1000) `div` 100) <> "s"
 
 
 -- | Single-line diagnostic for plain-text / shell output.
--- a/tricorder/src/Tricorder/TestOutput.hs
+++ b/tricorder/src/Tricorder/TestOutput.hs
@@ -1,4 +1,4 @@
-module Tricorder.TestOutput (parseHspecOutput, stripGhciNoise) where
+module Tricorder.TestOutput (parseHspecOutput, parseHspecDuration, stripGhciNoise) where
 
 import Data.Text qualified as T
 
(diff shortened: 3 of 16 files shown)


atelier/src/Atelier/Time.hs
```diff
@@ -45,3 +46,23 @@ instance FromJSON Minute where
...
+instance ToJSON Second where
+    toJSON s = toJSON (toMicroseconds s `div` 1_000_000)
+
+
+instance ToJSON Minute where
+    toJSON m = toJSON (toMicroseconds m `div` 60_000_000)
+
+
+instance ToJSON Hour where
+    toJSON h = toJSON (toMicroseconds h `div` 3_600_000_000)
```
Gipphe, line 68: question: These `ToJSON` instance are not the inverse of their corresponding `FromJSON` instances. Is this intentional?

Example:

```hs
decode @Millisecond (encode @Millisecond 200) == Nothing
```

tricorder/src/Tricorder/CLI/Render.hs
```diff
@@ -18,12 +19,13 @@ import Tricorder.SourceLookup (ModuleSourceResult (..))
...
-        show ms <> "ms"
-    else
-        show (ms `div` 1000) <> "." <> show ((ms `mod` 1000) `div` 100) <> "s"
+formatDuration :: Millisecond -> Text
+formatDuration d =
+    let ms = toMicroseconds d `div` 1000
+    in  if ms < 1000 then
+            show ms <> "ms"
+        else
+            show (ms `div` 1000) <> "." <> show ((ms `mod` 1000) `div` 100) <> "s"
```
Gipphe, line 28: nitpick (non-blocking): Prefer `where` if possible.

This is more of an opinion thing, but I'll use this example to bring up this point: do we prefer `where` or `let in`?

I personally prefer `where`, and try to use it instead of `let` whenever I can. We don't have to discuss this bike shed all that much, but what are your thoughts?

```suggestion
formatDuration d =
    if ms < 1000 then
        show ms <> "ms"
    else
        show (ms `div` 1000) <> "." <> show ((ms `mod` 1000) `div` 100) <> "s"
  where
    ms = ms = toMicroseconds d `div` 1000
```

tricorder/src/Tricorder/UI/Keys.hs
```diff
@@ -84,19 +94,23 @@ keys =
...
+        , ("cycle test view", CycleTestView)
         ]
 
 
 bindings :: [(KeyEvent, [Binding])]
 bindings =
     [ (ToggleDaemonInfoView, [bind 'g'])
-    , (Quit, [bind 'q', ctrl 'c', binding KEsc []])
+    , (Quit, [bind 'q', ctrl 'c'])
+    , (ExitView, [binding KEsc []])
```
Gipphe, line 110: suggestion: Add these new key events to the `README.md`'s "Custom Key Bindings" section.

tricorder/src/Tricorder/TestOutput.hs
```diff
@@ -1,36 +1,78 @@
...
+--
+-- Scans from the end of the line: strips the closing @)@, walks backwards
+-- through digits, @.@, @s@, @m@, @μ@, then requires @(@ followed by a space.
+-- Returns the input unchanged if the suffix does not match this pattern.
+stripTimingAnnotation :: Text -> Text
+stripTimingAnnotation t
+    | T.null t || T.last t /= ')' = t
+    | otherwise =
+        let withoutClose = T.init t
+            (timePart, rest) = T.span (\c -> isDigit c || c `elem` (".smμ" :: [Char])) (T.reverse withoutClose)
```
Gipphe, line 50: comment: I prophesize now that `.smµ` will cause a bug in the far future when someone working on a gargantuan project complains to us because this part is not parsed correctly for his test suite because it takes more than 1 hour to run 😛

This is not at all something we want to bother with at this point though, so it's fine as it is.

Gipphe (approved):
LGTM. Add the new key entries to `README`, and this is pretty much good to go.

tricorder/src/Tricorder/CLI/Render.hs
cgeorgii: I usually prefer `where` as well. :shrug:

tricorder/src/Tricorder/UI/Keys.hs
cgeorgii: Good point. Added.

atelier/src/Atelier/Time.hs
cgeorgii: No, definitely not. I am tempted to roll out our own time-related types, though, to avoid these orphan instances.

In any case, here's a ticket for the discrepant instances: https://github.com/atelier-hub/tricorder/issues/104

PR #105, opened 2026-05-21
feat: function-level source lookup via Module#function syntax

cgeorgii:
Extends `tricorder source` to accept an optional `#function` suffix:

- `Relude.Enum` — whole module source (existing behaviour)
- `Relude.Enum#inverseMap` — single top-level binding

**Implementation:**

- `SourceQuery` type in `GhcPkg.Types` carries the module name and an optional function name; replaces bare `ModuleName` throughout the protocol, CLI, and daemon
- `extractFunctionSource` splits the Haddock `<pre>` block on `<span id="line-N">` markers, locates the target binding's span, then expands backward/forward to the surrounding blank-line-delimited block (type sig + docstring above, pragmas below)
- `extractReExports` parses the export list to annotate whole-module results with re-exported names/modules; suppressed in function mode
- `stripAnnotations` removes `<span class="annottext">` hover-tooltip blocks (elaborated GHC types) before `stripTags`, preventing them from polluting the plain-text output; applied via the same line-chunk pipeline in both module and function mode
- `FunctionNotFound` result variant for when the module exists but the named binding is absent from the Haddock source HTML

### DIFF
--- a/tricorder/src/Tricorder/SourceLookup.hs
+++ b/tricorder/src/Tricorder/SourceLookup.hs
@@ -2,18 +2,24 @@ module Tricorder.SourceLookup
     ( -- * Types
       ModuleName
     , PackageId
+    , SourceQuery (..)
     , ModuleSourceResult (..)
+    , ReExport (..)
 
       -- * Lookup
     , lookupModuleSource
 
       -- * HTML extraction
     , extractSource
+    , extractFunctionSource
+    , extractReExports
+    , stripAnnotations
     , stripTags
     , unescapeEntities
     ) where
 
 import Data.Aeson (FromJSON, ToJSON)
+import Data.List (findIndex)
 import System.FilePath ((</>))
 
 import Data.ByteString.Lazy qualified as BSL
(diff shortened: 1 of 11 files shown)


tricorder/src/Tricorder/SourceLookup.hs
```diff
@@ -47,96 +65,213 @@ data ModuleSourceResult
...
+                        isModRe = T.isInfixOf "<span class=\"hs-keyword\">module</span>" before
+                        (_, rest3) = T.breakOn ">" afterClose
+                        innerHtml = fst $ T.breakOn "</a>" (T.drop 1 rest3)
+                        name = T.strip $ stripTags innerHtml
+                        entry
+                            | T.null name = Nothing
+                            | isLocal href = Nothing
+                            | isModRe = Just (ReExportModule name)
+                            | otherwise = Just (reExportName href name)
+                        rest' = T.drop (T.length "</a>") $ snd $ T.breakOn "</a>" rest
```
Gipphe, line 228: suggestion (non-blocking): Maybe more of a `todo:` form of comment, but would we be better served by using `megaparsec` to parse out the various parts of the source in these `extractX` functions?

Definitely not something to contend with in _this_ PR though.

Gipphe (approved):
Sweet stuff!

tricorder/src/Tricorder/SourceLookup.hs
cgeorgii: Definitely a direction worth exploring, especially since I've already done so for GHCi's output.

PR #107, opened 2026-05-22
refactor: replace semaphore-based restart with `restartableForkWith`

cgeorgii:
- Add `Conc.restartableForkWith`: a structured-concurrency primitive that runs a setup step in the scope before each fork, structurally guaranteeing the setup completes before the forked action starts
- Rewrite `withSession` and `withSubSession` to use it, eliminating the semaphore, the nested `withSession` fork, and the `Concurrent` constraint
- `withSubSession` now subscribes to `SessionStoreReloaded` directly via a recursive `awaitChange` filter, restarting only when the projected sub-session fields actually change


Gipphe (approved):
Love it! 🚀

PR #108, opened 2026-05-22
refactor: add Stream effect and eliminate State from withSubSession

cgeorgii:
- Add `Atelier.Effects.Stream`: an infinite pull-based stream backed by a buffered `Sub` subscription via `withStream`, with `map`, `filter`, and `changes` combinators
- Add `Conc.restartableForkLoop`: threads a value across restart iterations — the signal returns the next `r`, seeding the following fork directly
- Rewrite `withSubSession` using `Stream.withStream`, `Stream.changes`, and `restartableForkLoop`, eliminating internal `State subSession` entirely; the sub-session value is now threaded through the loop rather than stored
- `withStream` scopes the listener thread to the continuation, fixing the event subscription window between fork start and signal subscribe

### DIFF
--- a/atelier/src/Atelier/Effects/Stream.hs
+++ b/atelier/src/Atelier/Effects/Stream.hs
@@ -1,0 +1,52 @@
+module Atelier.Effects.Stream
+    ( Stream
+    , next
+    , fromEvents
+    , filter
+    , changes
+    ) where
+
+import Atelier.Effects.Chan (Chan)
+import Atelier.Effects.Conc (Conc)
+import Atelier.Effects.Publishing (Sub)
+import Prelude hiding (filter)
+
+import Atelier.Effects.Chan qualified as Chan
+import Atelier.Effects.Conc qualified as Conc
+import Atelier.Effects.Publishing qualified as Sub
+
+
+-- | An infinite pull-based stream of values.
+newtype Stream es a = Stream {next :: Eff es a}
+    deriving (Functor) via (Eff es)
+
+
+-- | Run a continuation with a buffered stream of 'Sub' events. The stream
+-- subscribes before the continuation runs, so no events are missed. The
+-- internal listener thread is scoped to the continuation.
+fromEvents
+    :: forall event es a
+     . (Chan :> es, Conc :> es, Sub event :> es)
+    => (Stream es event -> Eff es a)
+    -> Eff es a
+fromEvents use = Conc.scoped do
+    (inChan, outChan) <- Chan.newChan
+    Conc.fork_ $ Sub.listen_ (Chan.writeChan inChan)
+    use $ Stream (Chan.readChan outChan)
+
+
+filter :: (a -> Bool) -> Stream es a -> Stream es a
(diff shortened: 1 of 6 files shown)


atelier/src/Atelier/Effects/Stream.hs
```diff
@@ -0,0 +1,56 @@
...
+    => (Stream es event -> Eff es a)
+    -> Eff es a
+withStream use = Conc.scoped do
+    (inChan, outChan) <- Chan.newChan
+    Conc.fork_ $ Sub.listen_ (Chan.writeChan inChan)
+    use $ Stream (Chan.readChan outChan)
+
+
+map :: (a -> b) -> Stream es a -> Stream es b
+map f (Stream n) = Stream (f <$> n)
```
Gipphe: suggestion: Add `Functor` instance to `Stream`

```hs
deriving via (Eff es) instance Functor (Stream es)
```

That'll give us a more generalizable `map`.

atelier/src/Atelier/Effects/Stream.hs
```diff
@@ -0,0 +1,56 @@
...
+
+-- | Run a continuation with a buffered stream of 'Sub' events. The stream
+-- subscribes before the continuation runs, so no events are missed. The
+-- internal listener thread is scoped to the continuation.
+withStream
+    :: forall event es a
+     . (Chan :> es, Conc :> es, Sub event :> es)
+    => (Stream es event -> Eff es a)
+    -> Eff es a
+withStream use = Conc.scoped do
```
Gipphe: suggestion: This should have a name that more accurately reflects that it takes events, and exposes them as a stream.

Something like:

```suggestion
fromEvents
    :: forall event es a
     . (Chan :> es, Conc :> es, Sub event :> es)
    => (Stream es event -> Eff es a)
    -> Eff es a
fromEvents use = Conc.scoped do
```

So qualified use would look like:
```hs
Stream.fromEvents @SessionStoreReload \stream -> ...
```

Gipphe (approved):
A pull-based way of consuming events, I guess? Instead of `Sub.listen` pushing events onto its caller, the caller can `Stream.next` to get the next event.

Unsure if this is a proper "stream", really. It's got the consuming part of it, but to produce value in the stream you just capture events. Don't get me wrong: I think that's fine for our use-case for now, and it gives us room to expand organically and adapt `Stream` to how we want to use streams.

Side note:
I've been wanting to port Bluefin's `Stream` effect to Effectful and see if it is possible to do. Bluefin's `Stream` effect is however somewhat intricate and feature-rich, so I haven't given it a good enough go to really say whether it's viable or not. According to a talk by the creator of Bluefin, its `Stream` effect is not possible to implement in Effectful. I find that hard to believe, to be honest, but time will tell 😛

cgeorgii:
> Unsure if this is a proper "stream", really.

I think it is a proper stream but it's not full-featured - yet at least. Your concern might come from the fact that we only have the `fromEvents` constructor for streams but we could have others, e.g. `fromList`, etc.

PR #109, opened 2026-05-22
Allow package to be built on MacOS

qnikst:
This commit allows the package to be build on MacOS. Was tested locally using nix develop in tricorder package and using tricorder for another package


cgeorgii:
Thank you very much! I made a bot to update materialization for Mac going forward, so we now support it :fireworks:

PR #114, opened 2026-05-26
feat: dispatch :add/:unadd to GHCi from file events

cgeorgii:
Adds richer file events (`Added` / `Modified` / `Removed`) to the watcher and wires them through to the GHCi session so file creations/deletions update the loaded module set instead of always going through `:reload`.

The dispatcher uses the loaded-module map as the source of truth (populated from `:show modules`), so the FileEvent is just a hint — what we actually do depends on whether GHCi already knows the file:

| | Known to GHCi | Unknown |
|---|---|---|
| `Added` | `:reload` | `:add <path>` |
| `Modified` | `:reload` | `:add <path>` |
| `Removed` | `:unadd <module>` | no-op |

Unknown-Modified → `:add` covers the case where an editor writes a brand-new file that GHCi hasn't seen yet (otherwise `:reload` is a no-op for it). Unknown-Removed → no-op replaces the old fallback-to-reload.

Also fixes a pre-existing bug surfaced by testing this: a test suite that fails to compile (`cabal repl` opens, but `:main` errors with `Not in scope: 'main'`) was being silently reported as **passed** because `detectOutcome` only checked for `*** Exception: ExitFailure`. Now scans for `": error:"` in the captured output as a fallback signal:

```
test:tricorder-test  error: <interactive>:1:1: error: [GHC-76037] Not in scope: 'main'
```

### DIFF
--- a/atelier/src/Atelier/Effects/Debounce.hs
+++ b/atelier/src/Atelier/Effects/Debounce.hs
@@ -2,20 +2,23 @@
 --
 -- == Overview
 --
--- @debounced key action@ schedules @action@ to run after the settle window,
--- but cancels it if another call with the same @key@ arrives before the
--- window expires. This coalesces rapid bursts (e.g. two inotify events for
--- the same file) into a single callback.
+-- 'debouncedWith' is the fundamental operation: it accumulates rapid successive
+-- calls with the same key using a merge function, then fires the callback once
+-- after the settle window with the merged result.
+--
+-- 'debounced' is a convenience wrapper for the common case where only timing
+-- matters and the last-registered action should fire.
 --
 -- == Example
 --
 -- @
--- watchFilePaths watches \path ->
---     debounced path (markDirty (changeKindFor path))
+-- watchFilePaths watches \path event ->
+--     debouncedWith 100 mergeFileEvent path event (handleChange path)
 -- @
 module Atelier.Effects.Debounce
     ( -- * Effect
       Debounce
+    , debouncedWith
     , debounced
 
       -- * Interpreters
@@ -23,6 +26,7 @@ module Atelier.Effects.Debounce
     , runDebounceNoOp
     ) where
 
+import Data.Dynamic (Dynamic, fromDynamic, toDyn)
 import Effectful (Effect)
 import Effectful.Concurrent (Concurrent)
 import Effectful.Concurrent.STM (STM)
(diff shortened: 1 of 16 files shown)


atelier/src/Atelier/Effects/Debounce.hs
```diff
@@ -61,28 +81,33 @@ runDebounce
...
+                    mValue <- STM.atomically $ do
+                        current <- Map.lookup key state
+                        if fmap fst current == Just n then do
+                            Map.delete key state
+                            pure $ current >>= \(_, mDyn) -> mDyn >>= fromDynamic
                         else
-                            pure False
-                    when shouldFire (unlift action)
+                            pure Nothing
+                    for_ mValue (unlift . callback)
```
Gipphe, line 106: note (non-blocking): Technically, this can fail to call the callback for the last call to `debounceWith` if `fromDynamic` fails to convert the value from `Dynamic` to `value`, for whatever reason. This is effectively guaranteed to not happen for the last call to `debounceWith` for a given key though, so it's a bit of a moot point. 😉

atelier/src/Atelier/Effects/Debounce.hs
```diff
@@ -61,28 +81,33 @@ runDebounce
...
-                    Map.insert n key counters
+                    current <- Map.lookup key state
+                    let n = maybe 0 ((+ 1) . fst) current
+                    let prior = current >>= \(_, mDyn) -> mDyn >>= fromDynamic
+                    let merged = case prior of
+                            Just p -> merge p value
+                            Nothing -> value
+                    Map.insert (n, Just (toDyn merged)) key state
                     pure n
                 void $ fork $ do
```
Gipphe, line 97: question (non-blocking): I understand this is how it's always been so far, but won't this spin up a thread for _every_ call to `debounce` and each of those threads will inevitably have to wait out the fill `settleMs` delay before they realize they weren't needed anymore after all?

Thinking especially of file watching events where it might fire multiple times for each file change, or whenever you change git branch and a bunch of files change at the same time, the number of threads spawned might be significant for large codebases.

Instead of tackling this in this PR, I think it makes sense to leave this for a future PR though.

tricorder/test/Unit/Tricorder/BuilderSpec.hs
```diff
@@ -84,36 +85,83 @@ testRestartOnCabalChange = do
...
+                (results, _) <- runTest knownFoo distinctCtrls (SourceChangeDetected "/abs/path/Foo.hs" Modified)
+                results `shouldMatchList` [NewLoadResult epoch epoch reloadLr]
+
+        describe "when the file is not loaded in GHCi" do
+            it "calls controls.add (the editor just wrote a new file)" do
+                (results, _) <- runTest Map.empty distinctCtrls (SourceChangeDetected "/abs/path/New.hs" Modified)
+                results `shouldMatchList` [NewLoadResult epoch epoch addLr]
+
+    describe "Added" do
+        it "calls controls.add when the file is not loaded" do
```
Gipphe:
```suggestion
        describe "when the file is not loaded" $ it "calls controls.add" do
```

tricorder/test/Unit/Tricorder/BuilderSpec.hs
```diff
@@ -84,36 +85,83 @@ testRestartOnCabalChange = do
...
+            (results, _) <- runTest Map.empty distinctCtrls (SourceChangeDetected "/abs/path/Foo.hs" Added)
+            results `shouldMatchList` [NewLoadResult epoch epoch addLr]
+
+        it "calls controls.reload when the file is already loaded (re-add is a reload)" do
```
Gipphe:
```suggestion
        describe "when the file is already loaded" $ it "calls controls.reload (re-add is a reload)" do
```

tricorder/test/Unit/Tricorder/BuilderSpec.hs
```diff
@@ -84,36 +85,83 @@ testRestartOnCabalChange = do
...
+            (results, _) <- runTest knownFoo distinctCtrls (SourceChangeDetected "/abs/path/Foo.hs" Added)
+            results `shouldMatchList` [NewLoadResult epoch epoch reloadLr]
+
+    describe "Removed" do
+        it "calls controls.unadd when the file is loaded" do
```
Gipphe:
```suggestion
        describe "when the file is loaded" $ it "calls controls.unadd" do
```

Could sit under the same `describe` as the test above.

tricorder/test/Unit/Tricorder/BuilderSpec.hs
```diff
@@ -84,36 +85,83 @@ testRestartOnCabalChange = do
...
+            (results, _) <- runTest knownFoo distinctCtrls (SourceChangeDetected "/abs/path/Foo.hs" Removed)
+            results `shouldMatchList` [NewLoadResult epoch epoch unaddLr]
+
+        it "is a no-op when the file is not loaded" do
```
Gipphe:
```suggestion
        describe "when the file is not loaded" $ it "is a no-op" do
```

Could sit under the same `describe` as the first of these `Added` tests.

Gipphe (approved):

PR #118, opened 2026-05-26
refactor: use :show targets as source of truth for known files

cgeorgii:
The previous commit fixed the stale-results bug by always `:reload`-ing on Added/Modified events, which sidestepped the issue but lost the `:add` optimization for genuinely-new files. Replace that workaround with a proper fix to the underlying invariant.

GHCi has two views: `:show modules` lists only successfully-compiled modules, while `:show targets` lists the configured target set (populated by `cabal repl`) and survives failed compiles. The dispatcher must key on the latter to correctly identify "file GHCi knows but couldn't load this cycle" — the case that caused the regression.

- `LoadResult` gains `targetNames :: [Text]` from `:show targets`.
- `collectGhciResult` queries `:show targets` after `:show modules`.
- `Builder.resolveKnownTargets` joins this cycle's `:show modules` (definitive path↔name mapping) with prior state's carryover for any target that's missing from `:show modules`. Targets that have never compiled are dropped — the dispatcher's `unknown → :add` branch handles them.
- The original dispatch matrix is restored: `:add` is back for genuinely-new files, `:reload` for known files, `:unadd` for removed.
- Tests: `parseShowTargets` parser tests and `resolveKnownTargets` unit tests, including an explicit regression test for the failed-compile carryover.

### DIFF
--- a/tricorder/src/Tricorder/Builder.hs
+++ b/tricorder/src/Tricorder/Builder.hs
@@ -12,6 +12,7 @@ module Tricorder.Builder
     , handleInitialBuild
     , onRestart
     , reloadOnSourceChange
+    , resolveKnownTargets
     , setNewPhase
     ) where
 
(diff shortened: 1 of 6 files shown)


Gipphe (approved):
I have faith that you've got it this time 😛

PR #120, opened 2026-05-27
feat: abort in-flight tests on change with no waiters

cgeorgii:
When a source change arrives mid-build with no clients blocked in `waitUntilDone`, interrupt both the GHCi reload (existing behaviour) **and** the currently running test suite, then start a fresh cycle on the latest code instead of finishing the now-stale run.

- `TestRunner` gains `InterruptCurrent`, `ResetAbort`, `IsAborted`. The IO interpreter tracks the in-flight `GhciProcess` in a `TVar` and reuses the existing `interruptGhci` (process-group SIGINT) primitive.
- `Builder.interruptCurrent` (renamed from `interruptCurrentReload`) gates on `hasWaiters` and always fires `TestRunner.interruptCurrent` alongside `controls.interrupt`.
- `runTestsIfClean` returns `Maybe [TestRun]` — `Nothing` on abort. Caller skips the `Done` publish so no stale result reaches clients; the next change triggers a fresh `Building -> Testing -> Done` cycle.

### DIFF
--- a/tricorder/src/Tricorder/Builder.hs
+++ b/tricorder/src/Tricorder/Builder.hs
@@ -21,7 +21,7 @@ import Data.Time (diffUTCTime)
 import Effectful.Concurrent (Concurrent)
 import Effectful.Exception (bracket_, trySync)
 import Effectful.Reader.Static (Reader, ask)
-import Effectful.State.Static.Shared (State, execState, get, modify, put, state)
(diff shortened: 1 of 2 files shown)


Gipphe (approved):
A bit of a convoluted way of having a stateful aborting mechanism on first glance, but I fail to see a neater way to do it. Good work!

PR #130, opened 2026-05-28
fix: rebuild on every cabal change

cgeorgii:
`withSubSession` filters via the projected sub-session's `Eq`, so cabal-file edits that only change unprojected fields (ghc-options, build-depends) were silently dropped after hpack regenerated the cabal file.

Add `withReloadingSubSession` as a non-filtering sibling and have `withBuilderSession` use it instead.

### DIFF
--- a/tricorder/src/Tricorder/Effects/SessionStore.hs
+++ b/tricorder/src/Tricorder/Effects/SessionStore.hs
@@ -4,6 +4,7 @@ module Tricorder.Effects.SessionStore
     , get
     , withSession
     , withSubSession
+    , withReloadingSubSession
     , ActiveSession (..)
     , Reloader (..)
     , runSessionStore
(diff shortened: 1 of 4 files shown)


tricorder/src/Tricorder/Effects/SessionStore.hs
```diff
@@ -91,6 +92,34 @@ withSubSession mkSubSession initialSession action =
...
+     . ( Chan :> es
+       , Conc :> es
+       , SessionStore :> es
+       , Sub SessionStoreReloaded :> es
+       )
+    => (Session -> subSession)
+    -> Session
+    -> (Reloader es -> subSession -> Eff es a)
+    -> Eff es Void
+withReloadingSubSession mkSubSession initialSession action =
```
Gipphe, line 113: question: How does this function differ from `withSession`? The caller can just convert the `Session` into a `subSession` in their handler passed to `withSession`, so I don't really see how this function differs from `withSession`...

Gipphe (approved):
I feel like we both might have lost the plot a bit along the way here... In the past, we used the `Watcher` to detect `CabalChange`s, and restarted GHCi based on that. I messed around with the builder quite a lot, and I'm not sure whether that is the case any more. The `SessionStore` was primarily meant to be used to react to changes that require a restart, so `.tricorder.yaml` changes and `*.cabal` changes.

I originally made `SessionStore.withSession` because I wanted the provided `Session` to be tightly bound to a lifetime of some sort. I didn't want us to have to rely on an implicit relationship between a `SessionChanged` event and the consumers of `Session` having to `SessionStore.get` manually, since I knew one would follow the other in nearly every single case.

I just feel like we might be overcomplicating things in the `Builder` at this point. I presume your `Orchestrator` refactoring might help a ton here though, both in reducing state in `Builder` and `TestRunner`, and in making them more testable.

tricorder/src/Tricorder/Effects/SessionStore.hs
cgeorgii: This one doesn't match on whether the config changed or not - it always reloads the builder. Since we don't have 1:1 parity between the session and the values coming from cabal and ghci it's better to always reload then risk staleness.

Is it clearer now?

Gipphe: But `withSession` restarts the handler as well. It does so for all changes to `Session`, just like this `withReloadingSubSession`.

PR #136, opened 2026-05-29
refactor: replace SessionStore with events

Gipphe:
Each component should instead be responsible for figuring out what warrants a "restart" or not. Signal requests to restart through events instead of using continuations.

This PR introduces the `Restart` event and some mechanisms surrounding that as a clear pattern of `Restart` events emerged while refactoring this.

The `SessionStore.get` operations is entirely replaced by an `Input Session` effect, which is just backed by a `State Session` effect. The `SessionStore` component, which shared very little with the previous `SessionStore` effect, is responsible for handling requests to reload the session store.


Gipphe:
Redundant.

PR #182, opened 2026-06-10
fix: surface location-less GHCi load failures without false positives

cgeorgii:
## Summary

Makes a GHCi load that fails with a location-less error reliably surface as a build error. The motivating case is a home-unit GHC plugin that can't load under `cabal repl --enable-multi-repl`, which GHCi reports as `<no location info>: error:` with no source span and a `Failed,` summary.

## Changes

- **Single source of truth for load outcome.** `parseReload` now emits GHCi's terminal `Ok,`/`Failed,` summary as a `GSummary LoadOutcome` item. `reloadFailed` reads those items and trusts the **last** one — so output printed *during* a load (e.g. a Template Haskell splice line beginning `Failed,`) no longer triggers a spurious failure.
- **Synthetic failure when nothing is located.** `collectResult` appends a `<no location info>` error when GHCi reports `Failed,` with no located error, so the build can't read clean.
- **Failure survives watch-dir filtering.** `preserveFailureVisibility` re-attaches that synthetic failure after `filterToWatchDirs` when an error existed before filtering but none survived — covering the case where the only error lay outside the watched directories.
- **Location-less diagnostic handling.** `toDiagnostics` keeps location-less *errors* (genuine load failures) but drops location-less *warnings*; `mergeDiagnostics` clears location-less diagnostics each cycle so they can't persist forever; `filterToWatchDirs` keeps them (no path to test against a watch dir).
- **Tighter marker check.** `isLocationLess` now requires a closing `>`, so a real path merely starting with `<` (e.g. `<generated>/Foo.hs`) isn't mistaken for a GHCi marker.

### DIFF
--- a/tricorder/src/Tricorder/Builder/Dispatch.hs
+++ b/tricorder/src/Tricorder/Builder/Dispatch.hs
@@ -8,6 +8,7 @@ module Tricorder.Builder.Dispatch
     , fileMatchesAnyTarget
     , filterToWatchDirs
     , mergeDiagnostics
+    , preserveFailureVisibility
     ) where
 
 import Atelier.Effects.FileWatcher (FileEvent (..))
@@ -18,9 +19,13 @@ import Data.List qualified as List
 import Data.Map.Strict qualified as Map
 import Data.Set qualified as Set
 
-import Tricorder.BuildState (Diagnostic (..))
+import Tricorder.BuildState (Diagnostic (..), Severity (..))
 import Tricorder.Effects.GhciSession (LoadResult (..), LoadedModule (..))
-import Tricorder.Effects.GhciSession.GhciParser (pathSuffixesAsModuleName)
+import Tricorder.Effects.GhciSession.GhciParser
+    ( isLocationLess
+    , pathSuffixesAsModuleName
+    , unattributedFailure
+    )
 
 
 -- | The Builder's per-GHCi-session cache: what it last saw from GHCi plus its
@@ -57,9 +62,16 @@ type DiagnosticMap = Map FilePath [Diagnostic]
 -- by any new diagnostics produced for them in this cycle. Files absent from
 -- 'compiledFiles' were skipped by incremental compilation and retain their
 -- previous diagnostics unchanged.
+--
+-- Location-less diagnostics (see 'isLocationLess') are never keyed to a real
+-- source file, so they would never appear in 'compiledFiles' and would persist
+-- forever once raised. They describe the current load's outcome, so we clear
+-- them every cycle and let this cycle's 'diagnostics' re-add them if the
+-- failure is still present.
 mergeDiagnostics :: DiagnosticMap -> LoadResult -> DiagnosticMap
 mergeDiagnostics prev LoadResult {compiledFiles, diagnostics} =
-    let cleared = foldr Map.delete prev compiledFiles
+    let retained = Map.filterWithKey (\f _ -> not (isLocationLess f)) prev
+        cleared = foldr Map.delete retained compiledFiles
         newByFile = Map.fromListWith (++) [(d.file, [d]) | d <- diagnostics]
     in  Map.union newByFile cleared
 
@@ -141,10 +153,16 @@ dispatch knownTargets known fp event = case known of
 -- those with mangled filenames produced by the C preprocessor (e.g.
 -- @"In file included from ..."@) are dropped here, before they can enter the
 -- accumulation map where they would be impossible to evict.
+--
+-- Location-less diagnostics (see 'isLocationLess') are always kept: they carry
+-- no path to test against a watch dir, but they represent genuine build-level
+-- failures (e.g. a home-unit GHC plugin that can't load under
+-- @--enable-multi-repl@) that must not be dropped, or the build would silently
+-- read as clean.
 filterToWatchDirs :: FilePath -> [FilePath] -> [Diagnostic] -> [Diagnostic]
 filterToWatchDirs _ [] diags = diags
 filterToWatchDirs projectRoot watchDirs diags =
-    filter (isUnderAnyWatchDir . (.file)) diags
+    filter (\d -> isLocationLess d.file || isUnderAnyWatchDir d.file) diags
   where
     absWatchDirs = map toAbsWd watchDirs
     toAbsWd wd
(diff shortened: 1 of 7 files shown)


tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
```diff
@@ -488,6 +504,70 @@ collectResultCustom projectRoot loads modules targets =
...
+unattributedFailure =
+    BuildState.Diagnostic
+        { severity = SError
+        , file = "<no location info>"
+        , line = 0
+        , col = 0
+        , endLine = 0
+        , endCol = 0
+        , title = "GHCi reported a failed load with no located error"
+        , text = "GHCi reported a failed load with no located error.\nCheck the full GHCi output for details.\n"
```
Gipphe: suggestion: Direct the user where to find the full GHCi output (`tricorder log`).

tricorder/src/Tricorder/Builder/Dispatch.hs
```diff
@@ -158,3 +176,23 @@ filterToWatchDirs projectRoot watchDirs diags =
...
+-- load failed but every error it produced lay outside those dirs (e.g. a
+-- compile error in a sibling home unit not under @watchDirs@), filtering would
+-- leave no diagnostics and the broken build would look green. Detecting that an
+-- error was present /before/ filtering but none survived, we re-attach the
+-- location-less synthetic failure (which filtering always keeps) so the failure
+-- still surfaces.
+--
+-- Takes the pre-filter diagnostics and the post-filter diagnostics; returns the
+-- post-filter list, with the synthetic failure appended only when needed.
+preserveFailureVisibility :: [Diagnostic] -> [Diagnostic] -> [Diagnostic]
```
Gipphe, line 193: question: Do we ever want to _not_ use `preserveFailureVisibility` with `filterToWatchDirs`? 

If not, just have `filterToWatchDirs` use `preserveFailureVisibility`.

tricorder/test/Unit/Tricorder/Effects/GhciSession/GhciParserSpec.hs
```diff
@@ -311,3 +329,93 @@ testShowTargets = do
...
+--------------------------------------------------------------------------------
+-- collectResult: failed load with no located error
+--------------------------------------------------------------------------------
+
+-- 'collectResult' is the safety net: GHCi can end a load with @Failed, …@
+-- without emitting any error that carries a source span. The build must never
+-- read as clean in that case, so a synthetic error diagnostic is added.
+testUnattributedFailure :: Spec
+testUnattributedFailure = do
+    it "adds a synthetic error when the load failed but no error was located" do
```
Gipphe: nitpick:

```suggestion
    describe "when the load failed but no error was located" $ it "adds a synthetic error" do
```

tricorder/test/Unit/Tricorder/Effects/GhciSession/GhciParserSpec.hs
```diff
@@ -311,3 +329,93 @@ testShowTargets = do
...
+
+    it "adds nothing for a successful load" do
+        let reloadOutput =
+                [ "[1 of 1] Compiling Main ( app/Main.hs, interpreted )"
+                , "Ok, 1 module loaded."
+                ]
+            result = collectResult "/project" reloadOutput [] []
+        result.diagnostics `shouldBe` []
+
+    it "does not flag a clean build when 'Failed,' appears off the summary line" do
```
Gipphe: nitpick:

```suggestion
    describe "when 'Failed,' appears off the summary line" $ it "does not flag a clean build" do
```

tricorder/test/Unit/Tricorder/BuilderSpec.hs
```diff
@@ -753,6 +754,37 @@ testMergeDiagnostics = do
         let merged = mergeDiagnostics Map.empty result
         Map.lookup warnMsg.file merged `shouldBe` Just [warnMsg]
 
+    it "clears a stale location-less diagnostic when the cycle reports none" do
```
Gipphe: nitpick:

```suggestion
    describe "when the cycle reports none" $ it "clears a stale location-less diagnostic" do
```

tricorder/test/Unit/Tricorder/BuilderSpec.hs
```diff
@@ -794,6 +826,37 @@ testFilterToWatchDirs = do
             nixD = errMsg {file = "/nix/store/abc123/ghcautoconf.h"}
         filterToWatchDirs root ["."] [d, nixD] `shouldBe` [d]
 
+    it "keeps location-less <no location info> errors despite no path" do
```
Gipphe: nitpick:

```suggestion
    describe "when diagnostic has no path it" $ it "keeps location-less <no location info> errors" do
```

tricorder/test/Unit/Tricorder/BuilderSpec.hs
```diff
@@ -794,6 +826,37 @@ testFilterToWatchDirs = do
...
+        filterToWatchDirs root watchDirs [d] `shouldBe` [d]
+
+    it "does not treat a real <-prefixed path as a location-less marker" do
+        -- isLocationLess requires a closing '>'. A real (if exotic) path that
+        -- merely starts with '<' is an ordinary out-of-watch file and must be
+        -- dropped, not kept as a build-level marker.
+        let d = errMsg {file = "<generated>/Foo.hs"}
+        filterToWatchDirs root watchDirs [d] `shouldBe` []
+
+    it "a failed load does not read as clean when its only error is out of watch dirs" do
```
Gipphe: nitpick:

```suggestion
    describe "when its only error is out of watch dirs" $ it "a failed load does not read as clean" do
```

Gipphe (approved):
LGTM

Just got some questions and nits on spec naming 😛

tricorder/src/Tricorder/Effects/GhciSession/GhciParser.hs
cgeorgii: Added the ghci output to `tricorder log` and changed the message. Good call.

tricorder/src/Tricorder/Builder/Dispatch.hs
cgeorgii: Yes, the use of `filterToWatchDirs` in Builder:286 would be wrong with this - it would include this synthetic diagnostic. Perhaps we can change improve it so that this is not the case but right now it would be wrong.

PR #194, opened 2026-06-12
feat: display views as tabs

Gipphe:
Show active tab, and treat the individual views as tabs. Show key bind for each tab next to the tab name to make it more apparent what to press to view each tab.

This PR also drops the concept of a "navigation history" where `Esc` attempts to go back to the previous view that was opened. After some testing, I deemed it more confusing than helpful. If the user wants to go to a specific view they were at, it's easier for them to use the key binding that is shown instead of relying on `Esc` acting like a "back" button.

This also makes the Daemon info view a separate view that doesn't show the build result while it is open. The user is unlikely to need this daemon info while viewing the build result.

https://github.com/user-attachments/assets/cf0302c3-7701-4f6b-90f1-0efbf450172c

Depends on https://github.com/atelier-hub/tricorder/pull/196

### DIFF
--- a/tricorder/src/Tricorder/UI.hs
+++ b/tricorder/src/Tricorder/UI.hs
@@ -7,18 +7,11 @@ import Atelier.Effects.Conc (Conc)
 import Atelier.Effects.Console (Console)
 import Atelier.Effects.Delay (Delay)
 import Atelier.Effects.File (File)
-import Brick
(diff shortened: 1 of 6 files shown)


cgeorgii (approved):
The idea behind the view stack was more akin to modals than a back button. It doesn't matter, though, looks good. 👍

PR #196, opened 2026-06-12
chore: drop modules from materialized files

Gipphe:
Provide `package.yaml` files for the `haskell-plan-to-nix` stage of the builds so that the materialization uses `.cabal` files generated from `package.yaml` files in isolation, whereby `hpack` does not find any source modules to populate the `exposed-modules` list with.

This simplifies materialization quite a bit, since adding a module will no longer cause every set of materialized files to have to be re-materialized. Materialization is a bit cumbersome since we have it for multiple architectures, and multiple GHC versions.

Depends on https://github.com/atelier-hub/tricorder/pull/201

### DIFF
--- a/.github/workflows/update-materialization.yml
+++ b/.github/workflows/update-materialization.yml
@@ -93,11 +93,11 @@ jobs:
         if: ${{ steps.check-for-change.outputs.upload == 'true' }}
         with:
           name: materialization-${{ matrix.system }}-${{ matrix.ghc-version }}
-          # Include the README.md to make `upload-artifact` base the artifact
+          # Include the hie.yaml to make `upload-artifact` base the artifact
           # in the root of the repo.
           path: |
             nix/materialized/${{ matrix.system }}/${{ matrix.ghc-version }}
-            README.md
+            hie.yaml
           include-hidden-files: true
 
   commit:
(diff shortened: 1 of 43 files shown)


.github/workflows/update-materialization.yml
```diff
@@ -118,7 +118,6 @@ jobs:
       - uses: actions/download-artifact@v8
         with:
           pattern: materialization-*
-          path: nix/materialized
```
Gipphe, line 121: This was accidentally left in from the previous PR.

.github/workflows/update-materialization.yml
```diff
@@ -93,11 +93,11 @@ jobs:
         if: ${{ steps.check-for-change.outputs.upload == 'true' }}
         with:
           name: materialization-${{ matrix.system }}-${{ matrix.ghc-version }}
-          # Include the README.md to make `upload-artifact` base the artifact
+          # Include the hie.yaml to make `upload-artifact` base the artifact
```
Gipphe, line 96: `hie.yaml` is a considerably smaller file than `README.md`. Not that `README.md` is of a significant size, but since we have to pick a file in the root of the repo either way, we might as well pick the smallest one.

cgeorgii (approved):

PR #207, opened 2026-06-16
chore: add icon

Gipphe:
This PR consists of 4 distinct commits, each with a variation of an icon we can use to represent `tricorder`.

All of the variations are generated with the help of Claude.


cgeorgii (approved):

Gipphe:
Dropped since we can't think of an icon we like.

PR #210, opened 2026-06-16
chore: extract atelier into its own repository

cgeorgii:
Remove the in-repo atelier-prelude/core/db/testing packages and consume them from the pinned atelier source-repository-package instead. The flake now tracks `github:atelier-hub/atelier`, cabal.project lists only tricorder, and the dev shell builds the atelier libraries as ordinary dependencies.

Since `tricorder` is now the only package in this repository, flatten the repository by moving source from `tricorder/` to the root of the project.


Gipphe (changes requested):
In a way, it makes sense to keep `tricorder` separate from `atelier`, so I'm kinda _for_ making this split. But at the same time, we're doing it because `atelier-db` cannot be compiled with GHC 9.14, which is a bit of a weak reason. So for the sake of properly supporting GHC 9.14, we should instead opt to deal with `atelier-db` rather than split the entire repo.

cgeorgii:
Idea scrapped.

PR #235, opened 2026-06-22
refactor: read `cabal.project` and `.cabal` files only once

Gipphe:
Depends on https://github.com/atelier-hub/tricorder/pull/233

### DIFF
--- a/tricorder/src/Tricorder/Session.hs
+++ b/tricorder/src/Tricorder/Session.hs
@@ -15,6 +15,8 @@ module Tricorder.Session
     , TestTimeout (..)
     , Pattern
     , loadSession
+    , CabalFile (..)
+    , runCabalFiles
     , resolveCommand
     , resolveTargets
     , discoverCabalFiles
@@ -33,6 +35,7 @@ import Atelier.Types.WithDefaults (WithDefaults (..))
 import Data.Aeson (FromJSON (..), ToJSON (..))
 import Data.Default (Default (..))
 import Data.List (nub)
+import Data.Traversable (for)
 import Distribution.Compat.Lens (view)
 import Distribution.Fields (Field (..), FieldLine (..), Name (..), readFields)
 import Distribution.PackageDescription.Parsec (parseGenericPackageDescriptionMaybe)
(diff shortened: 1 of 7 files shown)


tricorder/src/Tricorder/Session.hs
```diff
@@ -470,9 +468,10 @@ cabalFilesIn dir = do
 -- skipped.
 projectPackageEntries :: ByteString -> [FilePath]
 projectPackageEntries contents =
-    case readFields contents of
-        Left _ -> []
-        Right fields -> filter (notElem '*') (concatMap fromField fields)
+    either (const []) id
+        $ filter (notElem '*')
+            . concatMap fromField
+            <$> readFields contents
```
cgeorgii: nitpick: I liked the previous one better in terms of readability.

tricorder/src/Tricorder/Session.hs
```diff
@@ -509,33 +508,61 @@ resolveTestTargets cfg targets = case cfg.testTargets of
...
+resolveWatchExclusionPatterns :: [Text] -> WatchExclusionPatterns
 resolveWatchExclusionPatterns rawPatterns = do
-    either (throwIO . userError . show) (pure . WatchExclusionPatterns)
-        $ traverse
-            ( parseRegex
-                . toString
-            )
-            rawPatterns
+    either (error . show) WatchExclusionPatterns
+        $ traverse (parseRegex . toString) rawPatterns
```
cgeorgii, line 513: I know the end effect is currently the same but we're making the error here _more_ implicit by making this function "pure". Let's consider introducing a proper error effect.

tricorder/src/Tricorder/Session.hs
```diff
@@ -509,33 +508,61 @@ resolveTestTargets cfg targets = case cfg.testTargets of
...
 
 
-loadSession
+data CabalFile = CabalFile
+    { projectFilePath :: FilePath
+    , projectPackageDescription :: GenericPackageDescription
```
cgeorgii, line 518: nitpick: we don't need the `project` prefix on each field here.

tricorder/src/Tricorder/Session.hs
Gipphe: Seeing it with fresh eyes after a few days, I think I agree.

tricorder/src/Tricorder/Session.hs
Gipphe: I agree. Made it more explicit by having it return an `Either`, and having `loadSession` deal with this `Either` properly.

cgeorgii (approved):
Nice :)

PR #238, opened 2026-06-23
fix: watch-dir scoping for bare package-name targets

### LINKED ISSUE #236: Incorrect rebuild trigger when specifying targets
With this configuration, touching a file in `tricorder/` triggers a rebuild:

```yaml
session:
  targets: [atelier-core]
```

This is incorrect since the touched file lives outside of the `atelier-core` file tree.

cgeorgii:
Closes #236 

## Summary

This branch fixes a watch-scoping bug and then - in a separate commit - refactors the surrounding code by replacing stringly-typed cabal targets with a small discriminated union.

The bug: with a config like `session: targets: [atelier-core]`, editing a file in a *sibling* package (e.g. `tricorder/`) triggered a rebuild, even though that file lives outside the `atelier-core` source tree. The cause was that a bare package-name target (no `lib:`/`exe:`/`test:` prefix) fell through `sourceDirsForTarget` to an empty catch-all, so target resolution produced no source dirs and `resolveWatchDirsFromTargets` fell back to `["."]` — watching the whole repo.

## What changed

- **Fix — scope watch dirs for bare package-name targets:** a bare target is now resolved as a package name (all of its components' source dirs) or as a single component matched by name. Added regression coverage for bare package/component names, the sibling-package rejection at the watcher layer, and the unparseable-cabal fallback.
- **Refactor — `Target` ADT:** introduced `Target` (`Qualified ComponentKind Text` / `Bare` / `Unrecognized`) with `parseTarget`/`renderTarget` sharing a single `kindPrefix` core via `inverseMap`, so the parser and printer can't drift apart. `sourceDirsForTarget` now matches on the ADT instead of `T.splitOn ":"` list shapes, and `Unrecognized` makes the old silent fall-through explicit.
- **`TestTargets` is now a guarded projection:** it wraps `[Target]` with an unexported data constructor, so the only ways to build one (`projectTestTargets` / `parseTestTargets`) filter to `test:` components — a `TestTargets` can never hold a non-test target. Both the `test_targets` config override and the derived path go through that projection, and the lib's `parseTestTargets` is reused by the tests rather than re-implemented.

## Behavior change

A non-`test:` entry in the `test_targets` config is now dropped rather than forwarded verbatim to the test runner. No existing behavior relied on this; it only affects misconfigured `test_targets`.

### DIFF
--- a/tricorder/src/Tricorder/Session.hs
+++ b/tricorder/src/Tricorder/Session.hs
@@ -33,13 +33,10 @@ import Atelier.Types.WithDefaults (WithDefaults (..))
 import Data.Aeson (FromJSON (..), ToJSON (..))
 import Data.Default (Default (..))
 import Data.List (nub)
+import Distribution.Compat.Lens (view)
 import Distribution.Fields (Field (..), FieldLine (..), Name (..), readFields)
 import Distribution.PackageDescription.Parsec (parseGenericPackageDescriptionMaybe)
-import Distribution.Types.Benchmark (benchmarkBuildInfo)
-import Distribution.Types.BuildInfo (hsSourceDirs)
 import Distribution.Types.CondTree (condTreeData)
-import Distribution.Types.Executable (buildInfo)
-import Distribution.Types.ForeignLib (foreignLibBuildInfo)
 import Distribution.Types.GenericPackageDescription
     ( GenericPackageDescription
     , condBenchmarks
@@ -50,11 +47,9 @@ import Distribution.Types.GenericPackageDescription
     , condTestSuites
     , packageDescription
     )
-import Distribution.Types.Library (libBuildInfo)
 import Distribution.Types.PackageDescription (package)
 import Distribution.Types.PackageId (pkgName)
 import Distribution.Types.PackageName (unPackageName)
-import Distribution.Types.TestSuite (testBuildInfo)
 import Distribution.Types.UnqualComponentName (mkUnqualComponentName, unUnqualComponentName)
 import Distribution.Utils.Path (getSymbolicPath)
 import Effectful.Exception (throwIO)
@@ -66,6 +61,7 @@ import Text.Regex.TDFA.ReadRegex (parseRegex)
 import Atelier.Effects.Log qualified as Log
 import Data.ByteString.Char8 qualified as BC
 import Data.Text qualified as T
+import Distribution.Types.BuildInfo.Lens qualified as Lens
 import Text.Regex.TDFA.Pattern qualified as Regex
 
 import Tricorder.Runtime (ProjectRoot (..))
@@ -348,18 +344,18 @@ sourceDirsForTarget gpd target =
         Unrecognized raw -> componentSourceDirsByName (T.takeWhileEnd (/= ':') raw)
   where
     mainPkgName = unPackageName . pkgName . package . packageDescription $ gpd
-    libDirs = hsSourceDirs . libBuildInfo
-    flibDirs = hsSourceDirs . foreignLibBuildInfo
-    testDirs = hsSourceDirs . testBuildInfo
-    exeDirs = hsSourceDirs . buildInfo
-    benchDirs = hsSourceDirs . benchmarkBuildInfo
 
-    mainLibSourceDirs = maybe [] (libDirs . condTreeData) (condLibrary gpd)
-    subLibSourceDirs name = dirsForComponent libDirs (condSubLibraries gpd) name
-    flibSourceDirs name = dirsForComponent flibDirs (condForeignLibs gpd) name
-    exeSourceDirs name = dirsForComponent exeDirs (condExecutables gpd) name
-    testSourceDirs name = dirsForComponent testDirs (condTestSuites gpd) name
-    benchSourceDirs name = dirsForComponent benchDirs (condBenchmarks gpd) name
+    -- @hs-source-dirs@ of any component, via the @HasBuildInfo@ lens — one
+    -- accessor that works uniformly across libraries, foreign libs, exes,
+    -- tests, and benchmarks, so we don't repeat a per-kind @buildInfo@ getter.
+    componentDirs component = view Lens.hsSourceDirs component
+
+    mainLibSourceDirs = maybe [] (componentDirs . condTreeData) (condLibrary gpd)
+    subLibSourceDirs name = dirsForComponent (condSubLibraries gpd) name
+    flibSourceDirs name = dirsForComponent (condForeignLibs gpd) name
+    exeSourceDirs name = dirsForComponent (condExecutables gpd) name
+    testSourceDirs name = dirsForComponent (condTestSuites gpd) name
+    benchSourceDirs name = dirsForComponent (condBenchmarks gpd) name
 
     -- Match a component name across every kind. The main library is keyed by
     -- the package name rather than an unqualified component name, so it joins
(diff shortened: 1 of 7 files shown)


tricorder/src/Tricorder/Session.hs
```diff
@@ -120,14 +125,92 @@ newtype Command = Command {getCommand :: Text}
...
+
+
+-- | Parse raw target strings (e.g. the @test_targets@ config) and project them
+-- onto their test suites — non-test entries are dropped.
+parseTestTargets :: [Text] -> TestTargets
+parseTestTargets = projectTestTargets . map parseTarget
+
+
+-- | A cabal build target, parsed from its textual @[kind:]name@ form. Used to
+-- resolve which source directories belong to a target.
```
Gipphe, line 154: question: Correct me if I'm wrong, but can't a target also be in the form `pkgName:kind:componentName`?

Gipphe: Turns out there can be 2 colons: `lib:foo:Data/ModuleName.hs` and `lib:foo:Data.ModuleName` are also valid.

See: https://hackage-content.haskell.org/package/Cabal-3.16.1.0/docs/src/Distribution.Simple.BuildTarget.html#UserBuildTargetTriple

tricorder/src/Tricorder/Session.hs
```diff
@@ -120,14 +125,92 @@ newtype Command = Command {getCommand :: Text}
...
+    | -- | A name with no @kind:@ prefix. Refers either to a package (all of its
+      -- components) or to a single component matched by name.
+      Bare Text
+    | -- | A form we don't recognize: an unknown kind, or extra colons.
+      Unrecognized Text
+    deriving stock (Eq, Show)
+
+
+-- | The kind of cabal component a 'Qualified' target names.
+data ComponentKind = Lib | Exe | Test
```
Gipphe: suggestion: If we are to go down this route of `ComponentKind`, we should support other forms of components such as `ForeignLib` and `Benchmark` as well.

If the user _really_ wants to specify a benchmark as a build target when using Tricorder, who are we to stop them?

tricorder/test/Unit/Tricorder/SessionSpec.hs
```diff
@@ -13,21 +13,7 @@ import Data.Map.Strict qualified as Map
...
-    , allComponentTargets
-    , compareTargets
-    , discoverCabalFiles
-    , resolveCommand
-    , resolveTargets
-    , resolveTestTargets
-    , resolveWatchDirs
-    , sourceDirsForTarget
-    )
+import Tricorder.Session (Command (..), ComponentKind (..), Config (..), Target (..), WatchDirs (..), allComponentTargets, compareTargets, discoverCabalFiles, parseTarget, parseTestTargets, resolveCommand, resolveTargets, resolveTestTargets, resolveWatchDirs, sourceDirsForTarget)
```
Gipphe, line 16: nitpick: How wide is your screen, man?! 😅

tricorder/test/Unit/Tricorder/SessionSpec.hs
```diff
@@ -37,58 +23,111 @@ spec_Session = do
...
+            parseTarget "lib:myapp-utils" `shouldBe` Qualified Lib "myapp-utils"
+
+        it "parses an exe: target" do
+            parseTarget "exe:myapp-exe" `shouldBe` Qualified Exe "myapp-exe"
+
+        it "parses a test: target" do
+            parseTarget "test:myapp-test" `shouldBe` Qualified Test "myapp-test"
+
+    describe "bare targets" do
+        it "parses a name with no kind prefix as bare" do
```
Gipphe: suggestion (non-blocking):

```suggestion
    describe "a name with no kind prefix" do
        it "parses as bare" do
```

tricorder/test/Unit/Tricorder/SessionSpec.hs
```diff
@@ -37,58 +23,111 @@ spec_Session = do
...
+
+-- | These exercise the 'Target' -> dirs resolution directly with constructed
+-- 'Target' values; the string -> 'Target' parsing is covered by 'testParseTarget'.
 testSourceDirsForTarget :: Spec
 testSourceDirsForTarget = do
-    describe "lib:" do
-        it "returns main library source dirs" do
-            sourceDirsForTarget gpd "lib:" `shouldBe` ["src"]
+    describe "Qualified Lib" do
+        context "when the name is empty" do
```
Gipphe, line 70: suggestion (non-blocking): Use `describe` instead of `context`.

`context` is an alias for `describe`. Might be wise to just stick with one or the other for consistency's sake.

Gipphe (approved):
LGTM, but it might make sense to include `Benchmark` and `ForeignLib` as `ComponentKind`s. Otherwise, `sourceDirsForTarget` does not return their directories as `WatchDirs`.

tricorder/test/Unit/Tricorder/SessionSpec.hs
cgeorgii: As discussed on Slack:

```hs
describe "some function" do
  context "when something is like this" do
    it "does that" do
      ...

  context "when something is like that" do
    it "does this" do
      ...
```
That's how I usually structure it and what I recommend: the top-level is a `describe`, `context` are variations, `it` are the actual tests.

tricorder/src/Tricorder/Session.hs
cgeorgii: Oh, I had no idea. :thinking: 

I'll check whether the unknown targets with 2 colons will pass through to the command correctly and if so I'll defer this with a ticket. Otherwise I'll fix it.

tricorder/src/Tricorder/Session.hs
```diff
@@ -223,61 +321,109 @@ resolveWatchDirsFromTargets cabalFiles (Targets targets) = do
...
-
--- | Infer the effective targets to build and watch.
--- Returns the configured targets as-is, or auto-detects all components across
--- every discovered package when no targets are configured.
-resolveTargets :: (FileSystem :> es) => [FilePath] -> [Text] -> Eff es Targets
-resolveTargets _ targets@(_ : _) = pure $ Targets $ sortBy compareTargets targets
+    benchDirs = hsSourceDirs . benchmarkBuildInfo
+
+    mainLibSourceDirs = maybe [] (libDirs . condTreeData) (condLibrary gpd)
+    subLibSourceDirs name = dirsForComponent libDirs (condSubLibraries gpd) name
```
Gipphe: suggestion (non-blocking): Use the [`hsSourceDir` lens](https://hackage-content.haskell.org/package/Cabal-syntax-3.16.1.0/docs/Distribution-Types-BuildInfo-Lens.html#v:hsSourceDirs):

```suggestion
import Distribution.Types.BuildInfo.Lens qualified as Lens
import Distribution.Compat.Lens (view)

    subLibSourceDirs name = dirsForComponent (view Lens.hsSourceDirs) (condSubLibraries gpd) name
```

Could thus also be included in `dirsForComponent`, to simplify even further.

Gipphe (approved):
Still LGTM 👌

tricorder/test/Unit/Tricorder/SessionSpec.hs
cgeorgii: Nah, that's a random picture from the internet, hehe :P

PR #239, opened 2026-06-25
fix: terminate whole cabal process group on shutdown

cgeorgii:
`cabal repl` is spawned in its own process group (`setCreateGroup True`) and forks build subprocesses (a worker `cabal`, `ghc`) that share it. The termination paths only signalled the group *leader* via `terminateProcess`, so those descendants were orphaned — the dangling `cabal` processes seen after `tricorder stop` and after a `.cabal`-triggered restart.

- Add `getProcessId` and `terminateProcessGroup` to the `Process` effect
- `stopGhciProcess`/`terminateGhciProcess` now capture the group id before the leader is reaped, then SIGTERM the whole group so no build subprocess is left behind. The pid must be read up front: once the leader exits (gracefully on `:quit` or via force-kill) and is reaped, the handle no longer yields a pid
- Add a regression test simulating the process tree deterministically

### DIFF
--- a/atelier-core/src/Atelier/Effects/Process.hs
+++ b/atelier-core/src/Atelier/Effects/Process.hs
@@ -33,17 +33,21 @@ module Atelier.Effects.Process
     , stopProcess
     , waitExitCode
     , interruptProcessGroup
+    , getProcessId
+    , terminateProcessGroup
 
       -- * Interpreters
     , runProcessIO
     ) where
 
+import Control.Exception (IOException, catch)
 import Effectful (Effect, IOE)
 import Effectful.Dispatch.Dynamic (interpret_)
 import Effectful.Exception (trySync)
 import Effectful.TH (makeEffect)
 import System.Exit (ExitCode (..))
-import System.Process (interruptProcessGroupOf)
+import System.Posix.Signals (sigTERM, signalProcessGroup)
+import System.Process (Pid, getPid, interruptProcessGroupOf)
 import System.Process.Typed
     ( ProcessConfig
     , createPipe
@@ -80,6 +84,18 @@ data Process :: Effect where
     -- | Send an interrupt (SIGINT) to the process's group. Requires the process
     -- to have been started with @'setCreateGroup' True@.
     InterruptProcessGroup :: TP.Process i o e -> Process m ()
+    -- | Look up the OS process id of a started process, or 'Nothing' if it has
+    -- already exited and been reaped. Capture this /before/ waiting on or
+    -- stopping the process if you need to address its group afterwards.
+    GetProcessId :: TP.Process i o e -> Process m (Maybe Pid)
+    -- | Send SIGTERM to an entire process group, identified by the group
+    -- leader's 'Pid'. For a process started with @'setCreateGroup' True@ the
+    -- leader's pid equals its process-group id, so passing its 'getProcessId'
+    -- result terminates the leader and every descendant sharing the group
+    -- (e.g. the build subprocesses @cabal repl@ forks) — not just the leader,
+    -- which is all @'stopProcess'@/@terminateProcess@ would reach. Errors (e.g.
+    -- the group already gone) are swallowed.
+    TerminateProcessGroup :: Pid -> Process m ()
 
 
 makeEffect ''Process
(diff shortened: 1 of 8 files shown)


tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
```diff
@@ -321,8 +324,18 @@ interruptGhci ghciProcess = do
...
+-- @cabal@, @ghc@) that share its group. 'Process.stopProcess' alone signals
+-- only the leader, orphaning those descendants as dangling processes.
 terminateGhciProcess :: (Process :> es) => GhciProcess -> Eff es ()
-terminateGhciProcess ghciProcess = Process.stopProcess ghciProcess.handle
+terminateGhciProcess ghciProcess = do
+    -- Capture the group id before 'stopProcess' reaps the leader (after which
+    -- 'getProcessId' returns 'Nothing'), then signal the whole group.
+    mpid <- getProcessId ghciProcess.handle
+    for_ mpid terminateProcessGroup
+    Process.stopProcess ghciProcess.handle
```
Gipphe: nitpick: Import this function in an unqualified manner to keep it consistent with other imports from `Atelier.Effects.Process`.

```suggestion
    stopProcess ghciProcess.handle
```

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
```diff
@@ -321,8 +324,18 @@ interruptGhci ghciProcess = do
...
+-- started with @'setCreateGroup' True@ and forks build subprocesses (a worker
+-- @cabal@, @ghc@) that share its group. 'Process.stopProcess' alone signals
+-- only the leader, orphaning those descendants as dangling processes.
 terminateGhciProcess :: (Process :> es) => GhciProcess -> Eff es ()
-terminateGhciProcess ghciProcess = Process.stopProcess ghciProcess.handle
+terminateGhciProcess ghciProcess = do
+    -- Capture the group id before 'stopProcess' reaps the leader (after which
+    -- 'getProcessId' returns 'Nothing'), then signal the whole group.
+    mpid <- getProcessId ghciProcess.handle
+    for_ mpid terminateProcessGroup
```
Gipphe: question: Won't `terminateProcessGroup` also terminate the leader?

Do we need the `Process.stopProcess` on the next line?

Gipphe (approved):
LGTM 💯 

Nice to get this fixed!

tricorder/src/Tricorder/Effects/GhciSession/GhciProcess.hs
cgeorgii: Nope, there's a gotcha here. Killing with `terminateProcessGroup` won't _reap_ a process, but `stopProcess` will.

This is a potential `Process` API improvement. I was going to create an issue for it but on second thought I think it's just better to do it now.

Gipphe: Right, gotta give it the ol' Linux `wait` to reap the process and free it from the process table. Good point!

atelier-core/src/Atelier/Effects/Process.hs
```diff
@@ -1,12 +1,13 @@
...
--- the one operation only @process@ provides (process-group interruption)
--- behind a single effect, so callers never depend on either package directly.
+-- Unifies @typed-process@ (configuration, lifecycle and stream capture) and the
+-- process-group signalling that @process@ provides behind a single effect, so
+-- callers never depend on either package directly.
 --
--- Build a 'ProcessConfig' with the re-exported @typed-process@ DSL
--- ('proc', 'shell', 'setStdin', …), then run it with one of the operations
--- below.
+-- Build a 'ProcessConfig' with the re-exported @typed-process@ DSL ('proc',
```
Gipphe, line 7: question: Shouldn't this comment be attached to `withProcessGroup` instead of sitting here at the top of the module?

atelier-core/src/Atelier/Effects/Process.hs
```diff
@@ -71,20 +73,77 @@ type RunningProcess = TP.Process
...
+    -- | Terminate the leader, wait for it to exit, and close its streams.
+    -- Internal: reaps and cleans up the leader but does /not/ reach the rest of
+    -- the group — 'withProcessGroup' / 'terminateProcessGroup' handle the group.
     StopProcess :: TP.Process i o e -> Process m ()
     -- | Block until the process exits and return its exit code.
     WaitExitCode :: TP.Process i o e -> Process m ExitCode
-    -- | Send an interrupt (SIGINT) to the process's group. Requires the process
-    -- to have been started with @'setCreateGroup' True@.
+    -- | Send SIGINT to the process's group. The process must have been started
+    -- with @'setCreateGroup' True@ (as 'withProcessGroup' does).
```
Gipphe: question (non-blocking): What happens if the process was started with `setCreateGroup False` and you call `InterruptProcessGroup` on it?

Just curious.

PR #252, opened 2026-06-29
fix: wait for daemon socket before declaring start done

cgeorgii:
`tricorder start` printed "Daemon started." as soon as the daemon was forked, before it had bound its Unix socket. A following command such as `tricorder status --wait` would then connect and crash with `Network.Socket.connect: does not exist (Connection refused)`.

- Add `isDaemonReady`, which probes actual socket connectability rather than process aliveness (the PID file appears before the socket binds, and may be missing right after the fork — so it is not a reliable readiness signal).
- `waitForDaemon` now polls the socket and returns whether it came up; `start` waits on it and reports accordingly.

### DIFF
--- a/tricorder/src/Tricorder/Daemon.hs
+++ b/tricorder/src/Tricorder/Daemon.hs
@@ -21,7 +21,7 @@ import Atelier.Effects.Posix.Daemons qualified as Daemons
 import Tricorder.Arguments (Force (..))
 import Tricorder.Effects.UnixSocket (UnixSocket)
 import Tricorder.Runtime (PidFile, SocketPath (..))
-import Tricorder.Socket.Client (isDaemonRunning, requestShutdown)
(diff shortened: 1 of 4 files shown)


tricorder/src/Tricorder/Daemon.hs
```diff
@@ -94,9 +94,31 @@ stopDaemon force = do
             pure ()
 
 
--- | Poll until the daemon socket becomes connectable.
-waitForDaemon :: (Daemons :> es, Delay :> es, Reader PidFile :> es, UnixSocket :> es) => Eff es ()
+-- | Poll until the daemon socket becomes connectable, giving up after roughly
```
Gipphe: nitpick: More correct wording.

```suggestion
-- | Poll until the daemon binds to the socket, giving up after roughly
```

Gipphe (approved):
LGTM 💯

PR #256, opened 2026-06-29
feat: add canvas flake template

cgeorgii:
Add a `canvas` flake template (`templates/canvas`) scaffolding an atelier + tricorder service: a Warp HTTP server, a rel8/Hasql database layer with sqitch migrations, YAML config loading, and Prometheus metrics. Wired into the repo as a flake template with CI verification builds (`canvas-hackage` against published atelier, `canvas-local` against in-repo sources).

- Encapsulate the template's check wiring in `nix/template.nix`, keeping `nix/outputs.nix` free of template-specific bindings.
- Decode the database section directly into atelier's `DBConfig` (requires atelier-db >= 0.1.1 for its `FromJSON` instance), dropping the bespoke wrapper type.
- Bump the `hackage` flake input so the template resolves atelier-db 0.1.1.0.


Gipphe (approved):
I'm not sure I would call this a "template", since templates are supposed to be a barebones starting point for a project, and thus are mostly incomplete on their own, but this "template" is closer to an "example repo", providing a full-fledged application as a starting point.

I think that's _more_ suitable in this case though, since `atelier` has been smelling a bit "frameworky" as of late, and this "template" or "example repo" of demonstrates all the various facets of `atelier` and its use quite well.

I will admit, I didn't go in-depth to see if the logic of the application makes a lot of sense, given that the application is just an example, but otherwise this LGTM 👍

PR #257, opened 2026-07-01
feat: restart the daemon from the TUI with R

cgeorgii:
Add a `restart_daemon` key event (bound to `R`) that restarts the daemon from `tricorder ui`, rebindable like the other key events.

- The brick event handler signals a worker thread over a channel, since it cannot run the daemon-control effects itself.
- The restart runs out-of-process by spawning `tricorder restart`: the daemon spawn double-forks, and forking the vty-controlled TUI process corrupts the terminal, so the fork must stay isolated.
- `queryWatch` stays patient while a restart we initiated is in flight, reconnecting to the fresh daemon instead of giving up.
- Extract a shared `restartDaemon` so the CLI `restart` subcommand and the TUI share one implementation.
- Expose `getExecutablePath` from `Atelier.Effects.Process` for re-invoking the program as a subprocess.
- Tie the README key-event list to the `KeyEvent` type with a tagref cross-reference.

### DIFF
--- a/tricorder/src/Tricorder/Socket/Client.hs
+++ b/tricorder/src/Tricorder/Socket/Client.hs
@@ -69,23 +69,30 @@ data Restarting = Restarting
 
 -- | Connect and stream build updates, calling the handler after each completed build.
 -- Retries automatically when the connection is lost or the daemon is restarting.
+--
+-- A transient drop is tolerated for a few attempts before giving up. While
+-- @isRestarting@ reports 'True' (a restart we initiated is in flight) the retry
+-- budget is ignored and reconnection continues indefinitely, so the watch
+-- survives the gap between stopping the old daemon and the new one binding its
+-- socket.
 queryWatch
     :: forall es
      . (Delay :> es, File :> es, UnixSocket :> es)
     => FilePath
+    -> Eff es Bool
+    -- ^ Whether a restart is currently in progress.
     -> (Either Restarting BuildState -> Eff es ())
     -> Eff es ()
-queryWatch sockPath handler = evalState retryLimit retryLoop
+queryWatch sockPath isRestarting handler = evalState retryLimit retryLoop
   where
     retryLimit = 3 :: Int
     retryLoop = do
         retries <- get @Int
-        if retries <= 0 then
-            pure ()
-        else do
+        keepGoing <- if retries > 0 then pure True else inject isRestarting
+        when keepGoing do
             void $ trySync $ withConnection sockPath \h -> sendQuery h Watch >> loop h
             Delay.wait (500 :: Millisecond)
-            modify $ subtract 1
+            when (retries > 0) $ modify $ subtract 1
             retryLoop
 
     loop h = do
--- a/tricorder/src/Tricorder/UI.hs
+++ b/tricorder/src/Tricorder/UI.hs
@@ -7,12 +7,18 @@ import Atelier.Effects.Conc (Conc)
 import Atelier.Effects.Console (Console)
 import Atelier.Effects.Delay (Delay)
 import Atelier.Effects.File (File)
+import Atelier.Effects.Process (Process, getExecutablePath, proc, readProcessStdout)
 import Brick (App (..), neverShowCursor)
+import Brick.BChan (BChan, writeBChanNonBlocking)
 import Brick.Keybindings (KeyConfig)
+import Data.IORef (IORef, newIORef, readIORef, writeIORef)
+import Effectful (IOE)
+import Effectful.Exception (trySync)
 import Effectful.Reader.Static (Reader, ask)
 
 import Atelier.Effects.Conc qualified as Conc
 
+import Tricorder.Daemon (waitForDaemon)
 import Tricorder.Effects.Brick (Brick)
 import Tricorder.Effects.BrickChan (BrickChan)
 import Tricorder.Effects.UnixSocket (UnixSocket)
@@ -30,7 +36,8 @@ import Tricorder.UI.State qualified as Model
 
 
 -- | Connect to the daemon and render a live-updating build status display using
--- a brick TUI. Quits on @q@ or @Esc@; arrow keys scroll the viewport.
+-- a brick TUI. Quits on @q@ or @Esc@; arrow keys scroll the viewport; @R@
+-- restarts the daemon.
 viewUi
     :: ( Brick :> es
        , BrickChan :> es
@@ -39,6 +46,8 @@ viewUi
        , Console :> es
        , Delay :> es
        , File :> es
+       , IOE :> es
+       , Process :> es
        , Reader Keys.Config :> es
        , Reader SocketPath :> es
        , UnixSocket :> es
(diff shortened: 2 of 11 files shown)


tricorder/src/Tricorder/UI/Keys.hs
```diff
@@ -111,6 +118,7 @@ bindings =
     [ (ToggleDaemonInfoView, [bind 'g'])
     , (ToggleHelp, [bind 'h'])
     , (CycleTestView, [bind 't'])
+    , (RestartDaemon, [bind 'R'])
```
Gipphe, line 121: nitpick: Keep in style with the other key bindings.

I don't think this impacts whether `SHIFT` needs to be pressed or not, but the other ones are specified in lower case, so 🤷 

```suggestion
    , (RestartDaemon, [bind 'r'])
```

cgeorgii: No, no, this is definitely not what I want. I want a capitalized `R` to prevent accidentally restarting the daemon, which could be "expensive" depending on the project.

Right now the daemon will only restart with `R`, not `r`.

tricorder/src/Tricorder/Socket/Client.hs
```diff
@@ -69,23 +69,30 @@ data Restarting = Restarting
...
         retries <- get @Int
-        if retries <= 0 then
-            pure ()
-        else do
+        keepGoing <- if retries > 0 then pure True else inject isRestarting
+        when keepGoing do
             void $ trySync $ withConnection sockPath \h -> sendQuery h Watch >> loop h
             Delay.wait (500 :: Millisecond)
-            modify $ subtract 1
+            when (retries > 0) $ modify $ subtract 1
```
Gipphe, line 95: nitpick: Unnecessary `when`

We can subtract the retries regardless.

```suggestion
            modify $ subtract 1
```

<hr>

thought: It's a bit overkill to use an entire state effect for this function, I think.

Since it's a recursive function, it could've just been 

```haskell
retryLoop retries = do
  keepGoing <- if retries > 0 then pure True else isRestarting
  when keepGoing do
    void $ trySync $ withConnection sockPath \h -> sendQuery h Watch >> loop h
    Delay.wait (500 :: Millisecond)
    retryLoop (retries - 1)
```

This drops the need to `inject isRestarting`, and it's just that little bit cleaner, imo.

tricorder/src/Tricorder/UI.hs
```diff
@@ -47,25 +56,62 @@ viewUi
...
+                (watchApp requestRestart keyConfig)
                 initialState
 
 
-watchApp :: KeyConfig KeyEvent -> App State Event Viewports
-watchApp kc =
+-- | Wait for restart requests from the TUI and service each one by spawning a
+-- fresh @tricorder restart@ process.
+--
+-- The restart must happen out-of-process: starting the daemon double-forks (see
```
Gipphe, line 83: comment: Sucks that the double-forking screws with Brick, complicating this drastically... I guess that's why this is not just a simple `stopDaemon >> startDaemon` solution.

tricorder/src/Tricorder/UI.hs
```diff
@@ -47,25 +56,62 @@ viewUi
...
+-- "System.Posix.Daemon"), and forking this vty-controlled process knocks the
+-- terminal out of the raw mode brick set up, so keystrokes stop being captured.
+-- Delegating to a separate process keeps the fork isolated. The brick event
+-- handler signals us over @restartChan@ because it cannot run these effects.
```
Gipphe, line 87: todo: We can use unlifting to run `Eff` actions in Brick's `EventM`, foregoing the need to use `IOE`, but that's an improvement to be made in a future PR.

tricorder/src/Tricorder/UI.hs
```diff
@@ -47,25 +56,62 @@ viewUi
 viewUi = do
     SocketPath sockPath <- ask
     chan <- BrickChan.newBChan 10
+    restartChan <- BrickChan.newBChan 1
```
Gipphe, line 59: question: Any particular reason to use Brick's `BChan` instead of `unagi-chan`?

tricorder/src/Tricorder/UI.hs
```diff
@@ -47,25 +56,62 @@ viewUi
...
+    => BChan ()
+    -> IORef Bool
+    -> Eff es ()
+restartWorker restartChan restarting = forever do
+    BrickChan.readBChan restartChan
+    liftIO $ writeIORef restarting True
+    self <- getExecutablePath
+    -- Swallow failures: a restart that errors must not take down the TUI, and
+    -- the flag has to be cleared either way.
+    _ <- trySync $ readProcessStdout $ proc self ["restart"]
```
Gipphe: nitpick: Since we're dropping the result, add a `readProcess` operation to `Process` and use that.

Doesn't impact the function of this line at all, but I assume `readProcess` skips a chunk of work that `readProcessStdout` would do that we ignore since we don't use the result.

```suggestion
    _ <- trySync $ runProcess $ proc self ["restart"]
```

tricorder/src/Tricorder/UI.hs
```diff
@@ -47,25 +56,62 @@ viewUi
...
+    void waitForDaemon
+    liftIO $ writeIORef restarting False
```
Gipphe: suggestion: Use `bracket_` to ensure the ref is set accordingly in the case of an exception in `waitForDaemon`, or anywhere else.


```suggestion
    bracket_
        (BrickChan.readBChan restartChan >> liftIO (writeIORef restarting True))
        (liftIO $ writeIORef restarting False)
        do
            self <- getExecutablePath
            -- Swallow failures: a restart that errors must not take down the TUI, and
            -- the flag has to be cleared either way.
            _ <- trySync $ readProcessStdout $ proc self ["restart"]
            void waitForDaemon

```

tricorder/src/Tricorder/UI.hs
cgeorgii: Yep, correct.

tricorder/src/Tricorder/UI/Keys.hs
Gipphe: I thought you had to add modifier keys explicitly. So you're telling me saying `'R'` here indicates `shift + r` correctly? Neat. I thought you had to specify it as `s-r`.

Gipphe (approved):
No glaring issues as far as I can see. LGTM 👍  Just some questions and minor suggestions/nits.

tricorder/src/Tricorder/Socket/Client.hs
cgeorgii: > thought: It's a bit overkill to use an entire state effect for this function, I think.

> Since it's a recursive function, it could've just been

I'm not sure this holds. If we use recursion like you suggested, it will track _total lifetime_ reconnect attempts. If the daemon restarts due to, say, a cabal change, this will count towards reconnect attempts and after 3 such instances the UI will give up retrying incorrectly.

We could reset the counter with explicit recursion but I think it will be noisier/fiddlier than `put retryLimit`.

tricorder/src/Tricorder/UI.hs
cgeorgii: None other than "it was there and it works". I went for what was simplest at the time and I think it's okay to keep it like this. Let me know if you feel strongly otherwise.

tricorder/src/Tricorder/UI.hs
cgeorgii: Great, well spotted :)

PR #260, opened 2026-07-03
feat: serve module source from sdist tarballs

cgeorgii:
Replace Haddock-HTML source extraction with reading a package's sdist tarball from cabal's global cache, so source is available even for packages built without documentation. The tarball is fetched on demand via `cabal fetch` when absent, and symbol queries are served by a pure lexical slicer.

- `SourceLookup.Tarball`: locate and read a module member in-process; search `CABAL_DIR`, XDG, and legacy `~/.cabal` cache layouts; prefer a module's library copy over any test/bench duplicate of the same name.
- `SourceLookup.Slice`: lexically slice the top-level declaration a symbol introduces (value, type, class, or constructor) together with its doc block — robust to CPP and to blank lines inside `where` clauses, and precise about operator, superclass, and constructor boundaries so neighbouring declarations are not swallowed.
- Distinguish a genuinely-absent package from a transient `cabal fetch` failure, caching only the former so a network blip is not pinned for the whole cache window.
- Drop the Haddock `getHaddockHtml`/`ReExport` machinery and the `SourceNoHaddock` result; add `tar`/`zlib` deps and `Env`/`Process` wiring for the new lookup path.


atelier-core/src/Atelier/Effects/Process.hs
```diff
@@ -154,6 +155,21 @@ readProcessSafe cmd args = do
...
+-- of @onRead@, and the process-lifecycle operations are unsupported. @onRead@
+-- runs in the remaining effects, so it can model a command's observable side
+-- effect — e.g. mutating a fake filesystem to mimic @cabal fetch@ populating a
+-- cache. Use this when the unit under test shells out solely via
+-- 'readProcessStdout'.
+runProcessReadWith
+    :: Eff es (ExitCode, LByteString)
+    -> Eff (Process : es) a
+    -> Eff es a
+runProcessReadWith onRead = interpret_ \case
```
Gipphe: suggestion: Unused. Remove this from this PR for now.

It seems very specialized to a particular test suites, so it should probably live in that test suite either way. It goes completely unused here for now though.

tricorder/src/Tricorder/Effects/Cabal.hs
```diff
@@ -0,0 +1,117 @@
...
+
+
+-- | Whether an on-demand fetch exited cleanly. A clean exit that still leaves
+-- no tarball is a /deterministic/ absence (safe to cache); a failed fetch is
+-- /transient/ (must not be cached).
+data FetchResult = Fetched | FetchFailed
+    deriving stock (Eq, Show)
+
+
+data Cabal :: Effect where
```
Gipphe, line 50: suggestion: Note here that this effect is specifically for using _the user's cabal_ in their repo.

I was about to ask "did you try using `Cabal`'s package from Haskell?", but then I remembered that we want to use the user's own `cabal` executable, not bring our own as a Haskell dependency.

tricorder/src/Tricorder/Effects/Cabal.hs
```diff
@@ -0,0 +1,117 @@
...
+    => Eff (Cabal : es) a
+    -> Eff es a
+runCabalIO = interpret \_ -> \case
+    FetchSource pkgId -> do
+        ProjectRoot projectRoot <- ask
+        Log.info $ "Source: cabal fetch " <> unPackageId pkgId
+        let cfg =
+                setWorkingDir projectRoot
+                    $ proc "cabal" ["fetch", "--no-dependencies", toString (unPackageId pkgId)]
+        result <- trySync (readProcessStdout cfg)
```
Gipphe: suggestion (non-blocking): Read both `stdout` and `stderr` to provide a better error message.

tricorder/src/Tricorder/Effects/Cabal.hs
```diff
@@ -0,0 +1,117 @@
...
+-- | Test interpreter: every 'fetchSource' yields the result of @onFetch@, which
+-- runs in the remaining effects so it can model the fetch's observable side
+-- effect — e.g. populating a fake filesystem to mimic a warmed cache.
+-- 'cabalVersion' is unsupported. Use this when the unit under test drives
+-- @cabal@ solely through 'fetchSource'. Mirrors
+-- 'Atelier.Effects.Process.runProcessReadWith'.
+runCabalFetchWith :: Eff es FetchResult -> Eff (Cabal : es) a -> Eff es a
+runCabalFetchWith onFetch = interpret \_ -> \case
+    FetchSource _ -> onFetch
+    CabalVersion -> error "runCabalFetchWith: cabalVersion is not supported"
```
Gipphe, line 101: nitpick: Couldn't you just provide `Nothing` here?

```suggestion
    CabalVersion -> pure Nothing
```

tricorder/src/Tricorder/SourceLookup/Slice.hs
```diff
@@ -0,0 +1,275 @@
...
+
+
+-- | Slice the declaration that introduces @symbol@ from @source@.
+--
+-- Casing of @symbol@ selects the search. An uppercase initial looks for a
+-- type-level declaration (@data@ \/ @newtype@ \/ @type@ \/ @type family@ \/
+-- @data family@ \/ @class@) and, failing that, the @data@ \/ @newtype@ block
+-- that defines it as a constructor. A lowercase or symbolic initial looks for a
+-- value binding (including operators in @(op)@ form). Returns 'Nothing' when no
+-- introducing declaration is found.
```
Gipphe: suggestion: Drop this entire paragraph.

Not only does this paragraph talk about implementation details, it talks about _other functions'_ implementation details.

tricorder/src/Tricorder/SourceLookup/Slice.hs
```diff
@@ -0,0 +1,275 @@
...
+                sliceType symbol ls <|> sliceConstructor symbol ls
+            else
+                sliceValue symbol ls
+
+
+-- | Does the query name a type-level entity? Decided purely by the first
+-- character: uppercase means a type, class, or constructor; anything else
+-- (lowercase or an operator symbol) means a value binding.
+isTypeQuery :: Text -> Bool
+isTypeQuery = maybe False (isUpper . fst) . T.uncons
```
Gipphe: suggestion: Rename this and drop the implementation details from the haddock.

The requested symbol is not quite a "query", if you ask me. It is a specifier, or merely a name.

The comment also is an English word-for-word translation of the function body.

```suggestion
-- | Evaluate whether the symbol references a type-level entity.
isTypeSymbol :: Text -> Bool
isTypeSymbol = maybe False (isUpper . fst) . T.uncons
```

tricorder/src/Tricorder/SourceLookup/Slice.hs
```diff
@@ -0,0 +1,275 @@
...
+
+
+-- ── Value bindings ─────────────────────────────────────────────────────────
+
+-- | Slice the value binding introducing @name@ (a function, CAF, or operator).
```
Gipphe, line 43: praise: I just want to highlight that _this_ is a good comment!

I feel like I'm a bit of a downer with how harsh I'm being towards Claude's comments... Then again, I guess I shouldn't feel too bad about the work of a machine 😅

tricorder/src/Tricorder/SourceLookup/Slice.hs
```diff
@@ -0,0 +1,275 @@
...
+
+
+-- | Does the query name a type-level entity? Decided purely by the first
+-- character: uppercase means a type, class, or constructor; anything else
+-- (lowercase or an operator symbol) means a value binding.
+isTypeQuery :: Text -> Bool
+isTypeQuery = maybe False (isUpper . fst) . T.uncons
+
+
+-- ── Value bindings ─────────────────────────────────────────────────────────
```
Gipphe, line 41: thought: Claude just _cannot_ stay consistent on the section heading styles it uses...

In one place it might use

```
-- ── Value bindings ─────────────────────────────────────────────────────────
```

And in another place, it'll use

```
------------------------------------------------------------------------------
-- Value bindings
------------------------------------------------------------------------------
```

This is however not a hill I'm willing to even stand on, let along _die_ on, so it doesn't matter much to me as long as we're consistent within a single module.

tricorder/src/Tricorder/SourceLookup/Slice.hs
```diff
@@ -0,0 +1,275 @@
...
+-- | Whether a column-0 line is a declaration head for the type-level @name@.
+introducesType :: Text -> Text -> Bool
+introducesType name line =
+    isCol0 line
+        && case T.words line of
+            ("data" : "family" : n : _) -> tok n == name
+            ("type" : "family" : n : _) -> tok n == name
+            ("data" : n : _) -> tok n == name
+            ("newtype" : n : _) -> tok n == name
+            ("type" : n : _) -> tok n == name
```
Gipphe, line 105: thought: This catches role annotations like `type role Foo` incorrectly.

I guess it's not a problem in this particular case though, since that'd just use `role` as `n`, and find that `tok "role" == name` is `False` either way. And the `type role` annotation doesn't introduce a type by itself.

tricorder/src/Tricorder/SourceLookup/Slice.hs
Gipphe, line 1: thought: This is a lot of stuff for parsing out one symbol in a Haskell file.

It might cause some trauma for me to ask this, but did you consider using GHC's parser for this? I don't know how ergonomic [GHC's parser](https://hackage-content.haskell.org/package/ghc-9.14.1/docs/GHC-Parser.html) is to use directly, but there are wrappers around it to make it saner to use, like [ghc-parser](https://hackage.haskell.org/package/ghc-parser). GHC's parser even supports parsing out Haddock comments.

tricorder/src/Tricorder/SourceLookup/Slice.hs
```diff
@@ -0,0 +1,275 @@
+-- | A pure, lexical source slicer.
+--
+-- Given a module's raw source text and a symbol name, 'sliceSymbol' returns the
+-- top-level declaration that introduces that symbol: the declaration head, any
+-- type signature directly above it, and the contiguous doc-comment block above
+-- that — terminating at the blank line or the next top-level declaration that
+-- ends it.
```
Gipphe: suggestion: Move this paragraph to `sliceSymbol`.

It is a nice high-level description of what `sliceSymbol` will return, but it is misplaced here at the module-level.

tricorder/src/Tricorder/SourceLookup/Tarball.hs
```diff
@@ -0,0 +1,255 @@
...
+-- in cabal's global package cache.
+--
+-- The cache holds one @\<pkg\>-\<ver\>.tar.gz@ per resolved package at a
+-- predictable path. On a hit we read a single module member in-process (no
+-- network, no extraction to disk). On a miss we warm the cache with
+-- @cabal fetch --no-dependencies@ — the exact version @ghc-pkg@ reports — and
+-- read the now-present tarball. Every step degrades to 'Nothing' rather than
+-- throwing, so a missing index, an offline machine, a yanked version, or a
+-- corrupt archive surfaces as "no source", never a crash.
+module Tricorder.SourceLookup.Tarball
```
Gipphe: suggestion: Move this to `obtainTarball`'s Haddock.

This paragraph specifically describes the `obtainTarball` function, not the module as a whole.

tricorder/src/Tricorder/SourceLookup/Tarball.hs
```diff
@@ -0,0 +1,255 @@
...
+hackageRepo :: FilePath
+hackageRepo = "hackage.haskell.org"
+
+
+-- ── High-level ─────────────────────────────────────────────────────────────
+
+-- | The result of locating (and, if needed, fetching) a package's tarball.
+data TarballOutcome
+    = -- | The tarball is present at this path.
+      TarballAt FilePath
```
Gipphe: nitpick: Redundant comment. Remove.

```suggestion
    = TarballAt FilePath
```

tricorder/src/Tricorder/SourceLookup/Tarball.hs
```diff
@@ -0,0 +1,255 @@
...
+    | -- | No tarball, but the lookup completed cleanly: a deterministic negative
+      -- (the package is not on any configured repository, or is yanked). Safe to
+      -- cache.
```
Gipphe: nitpick: This is an overly convoluted way to just say "we had a cache miss"...

```suggestion
    | -- | Tarball not found in cache.
```

tricorder/src/Tricorder/SourceLookup/Tarball.hs
```diff
@@ -0,0 +1,255 @@
...
+      TarballAbsent
+    | -- | The on-demand @cabal fetch@ itself failed (offline, stale index, …): a
+      -- /transient/ negative that must not be cached, lest a brief network blip
+      -- pin unavailability for the whole cache window.
```
Gipphe: suggestion: Reconsider the necessity of this comment as a whole.

I feel the name `TarballFetchFailed` is very descriptive as-is. If we _shouldn't_ cache for this sort of result, that's something to document _where caching happens_.

tricorder/src/Tricorder/SourceLookup/Tarball.hs
```diff
@@ -0,0 +1,255 @@
...
+      TarballFetchFailed
+    deriving stock (Eq, Show)
+
+
+-- | Locate @pkgId@'s sdist tarball in the cabal cache, fetching it on demand if
+-- absent. Distinguishes a genuinely-absent package from a failed fetch so the
+-- caller can decide what is cacheable.
```
Gipphe: Related to my comment on `TarballFetchFailed` above.

> Distinguishes a genuinely-absent package from a failed fetch so the caller can decide what is cacheable.

Even here, the comment admits that it's up to the caller to decide what to do with the result, further proving that it's not really the responsibility of `TarballOutcome` and its Haddock to determine what should be cached or not.

tricorder/src/Tricorder/SourceLookup/Tarball.hs
```diff
@@ -0,0 +1,255 @@
...
+--
+-- The fetch runs via the 'Cabal' effect, whose interpreter pins @cabal fetch@
+-- to the project root so it picks up the project's configured repositories
+-- (e.g. CHaP) and constraints.
```
Gipphe: nitpick: Remove this paragraph, or move it to `fetchSource`'s Haddock.

Implementation details about functions this function uses do not belong in this function's Haddock, unless they pertain specifically to the contract between this function and its caller.

tricorder/src/Tricorder/SourceLookup/Tarball.hs
```diff
@@ -0,0 +1,255 @@
...
+
+
+-- ── Locate ─────────────────────────────────────────────────────────────────
+
+-- | Search every candidate cabal cache directory (and every repository subdir
+-- within it) for @pkgId@'s tarball, preferring @hackage.haskell.org@.
+findTarball :: (Env :> es, FileSystem :> es) => PackageId -> Eff es (Maybe FilePath)
+findTarball pkgId = do
+    env <- getEnvironment
+    candidates <- concat <$> mapM basePaths (cabalPackagesDirs env)
```
Gipphe: nitpick: If we don't need the stronger `Monad` proof, use `traverse` to stick with just `Applicative`.

```suggestion
    candidates <- concat <$> traverse basePaths (cabalPackagesDirs env)
```

tricorder/src/Tricorder/SourceLookup.hs
```diff
@@ -63,234 +49,86 @@ data ModuleSourceResult
...
-extractReExports :: ModuleName -> Text -> [ReExport]
-extractReExports modName html =
-    let afterMod = snd $ T.breakOn "<span class=\"hs-keyword\">module</span>" html
-        exportRgn = fst $ T.breakOn "<span class=\"hs-keyword\">where</span>" afterMod
-    in  go exportRgn
+                    pure result
+                Nothing -> serveFromTarball query p
+
+
+-- | Resolve a module to its package, consulting the module→package cache first.
```
Gipphe: nitpick: This arrow did _not_ render correct for me on GitHub... I suggest replacing it with a simpler one.

```suggestion
-- | Resolve a module to its package, consulting the module -> package cache first.
```

tricorder/test/Unit/Tricorder/SourceLookup/SliceSpec.hs
Gipphe, line 1: suggestion: Add multi-line haddock comments case to all these tests.

I see you have one in `capturePrecision`, but you're testing different functions over each of these `describe` blocks, and most of them do not have a test case for multi-line haddock comments, which are quite common.

Gipphe (approved):
Overall, this is looking really good!

Some suggestions on function comments and a few nitpicks here and there, but nothing that'd stop this PR as-is if you disagree with my comments 👍 

Also, I just want to clarify: comments marked `nitpick:` are _completely_ optional if you disagree with them in any way. You can literally disregard _all_ of them and I wouldn't think differently about neither you nor the PR in question. If I find a suggestion to be more important, it'd be a `suggestion (non-blocking):` (or even a `suggestion (blocking):` if I'm feeling extra spicy) 😉

tricorder/src/Tricorder/Effects/Cabal.hs
cgeorgii: It's just a test function, better to fail than be surprised.

tricorder/src/Tricorder/SourceLookup/Slice.hs
cgeorgii: Yes, but it would couple us to a specific version of GHC. Similar line of thought to not using the cabal _library_ as you pointed above.

PR #261, opened 2026-07-03
refactor: add PostBuild

Gipphe:
Once a build is done, there can be a multitude of other things that take
place with the finalized result. One of these things is testing, but in
the future we will also support eval comments that have to be evaluated,
and potentially other "post-build" processes. Keeping these post-build
phases as proper `BuildPhase` values restricts the order they can be
performed, and duplicates the `BuildResult` across multiple post-build
states.

This PR renames `Done` to `BuildComplete` to better signal that it's the
GHC build that is complete, while tests and other post-build processes'
states are kept in `PostBuild` to be updated later by their respective
processes.

### DIFF
--- a/tricorder/test/Unit/Tricorder/BuilderSpec.hs
+++ b/tricorder/test/Unit/Tricorder/BuilderSpec.hs
@@ -32,19 +32,7 @@ import Control.Concurrent.STM qualified as STM
 import Data.Map.Strict qualified as Map
 import Data.Set qualified as Set
 
-import Tricorder.BuildState
-    ( BuildId (..)
-    , BuildPhase (..)
-    , BuildResult (..)
-    , BuildState (..)
-    , CabalChangeDetected (..)
-    , DaemonInfo (..)
-    , Diagnostic (..)
-    , Severity (..)
-    , SourceChangeDetected (..)
-    , TestRun (..)
-    , TestRunCompletion (..)
-    )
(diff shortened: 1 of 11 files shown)


tricorder/test/Unit/Tricorder/BuilderSpec.hs
```diff
@@ -32,19 +32,7 @@ import Control.Concurrent.STM qualified as STM
...
-    , BuildState (..)
-    , CabalChangeDetected (..)
-    , DaemonInfo (..)
-    , Diagnostic (..)
-    , Severity (..)
-    , SourceChangeDetected (..)
-    , TestRun (..)
-    , TestRunCompletion (..)
-    )
+import Tricorder.BuildState (BuildId (..), BuildPhase (..), BuildResult (..), BuildState (..), CabalChangeDetected (..), DaemonInfo (..), Diagnostic (..), PostBuild (..), Severity (..), SourceChangeDetected (..), TestPhase (..), TestRun (..), TestRunCompletion (..))
```
cgeorgii, line 35: Something something wide screen :P

cgeorgii (approved):

tricorder/test/Unit/Tricorder/BuilderSpec.hs
Gipphe: Gawd dammit...

It's the code actions, y'know... There is probably a `fourmolu` rule to remedy this.

PR #262, opened 2026-07-03
docs: document features

Gipphe:
Depends on https://github.com/atelier-hub/tricorder/pull/269


cgeorgii (commented):
Good idea but let's wait until eval comments are actually supported.

PR #263, opened 2026-07-06
refactor: separate actions made to the build result after build

Gipphe:
Tests and (the upcoming) eval comments perform updates on the build
result _after_ the build has completed. Instead of having to work around
the various possible states of `BuildState`, `PostBuildStore` provides
an effect that can only be interpreted with a proper `PostBuild` value
from a completed build.

### DIFF
--- a/tricorder/src/Tricorder/Builder.hs
+++ b/tricorder/src/Tricorder/Builder.hs
@@ -47,7 +47,6 @@ import Tricorder.BuildState
     , BuildResult (..)
     , CabalChangeDetected (..)
     , Diagnostic (..)
-    , PostBuild (..)
     , Severity (..)
     , SourceChangeDetected (..)
     , TestPhase (..)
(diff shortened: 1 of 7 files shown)


cgeorgii:
The mechanism (a new effect that's run directly during a build) seems a bit heavyweight. Unless I'm missing something - maybe the eval comment feature justifies it? - couldn't we achieve the same with helpers like these:

```hs
enterPostBuild :: BuildStore :> es => PostBuild -> Eff es ()
enterPostBuild pb = modifyPhase \_ -> BuildComplete pb

modifyPostBuild :: BuildStore :> es => (PostBuild -> PostBuild) -> Eff es ()
modifyPostBuild f = modifyPhase \s -> case s.phase of
    BuildComplete pb -> BuildComplete (f pb)
    other            -> other
```

We would lose the ability to mock the `PostBuildStore` effect seam but I don't see how that's useful at the moment.

Gipphe:
Something like `modifyPostBuild` is what I initially came up with, but I wanted stronger guarantees from the perspective of `requestTestRunsForNewBuildResults` and `runEvalsIfClean`.

Testing and eval comments are probably only the beginning of the various "post-build" tasks that Tricorder will be performing in the future.

`PostBuildStore` is an effect for tasks that should _only_ perform operations on the build state _after a build_. With just `PostBuildStore`, instead of `BuildStore`, they cannot set the `BuildState` to something like `Building Nothing`, for example.

`PostBuildStore` helps clearly distinguish what the post-build tasks can and will do. `BuildStore` has grown a bit hairy, if you ask me, with a lot of spurious operations that might warrant splitting out into separate effects, both for testability but also to better distinguish what each function does to the current `BuildState`. `PostBuildStore` is a small, initial step in that direction.

cgeorgii:
> BuildStore has grown a bit hairy, if you ask me, with a lot of spurious operations that might warrant splitting out into separate effects

I agree with this but the approach in this PR is not really splitting an effect out, but rather laying a new effect on top of the already complicated `BuildStore`.

One thought that could maybe solve the "actions made to the build results after build" issue the PR is trying to solve: what if we move the test results out of the `BuildResult` and into the `PostBuild` structure? Then you'd have a guarantee that updating test results wouldn't be able to touch build state.

On top of that, we could move the `BuildResult` into the `BuildPhase` so it's guaranteed not to be changed when assembling the PostBuild. Something like this:

```hs

data BuildPhase
    = Building (Maybe BuildProgress)
    | Restarting
    | BuildComplete BuildResult PostBuild   -- frozen result + accruing post-build state, side by side
    | BuildFailed Text

data BuildResult = BuildResult
    { completedAt :: UTCTime
    , duration    :: Millisecond
    , moduleCount :: Int
    , diagnostics :: [Diagnostic]
    }   -- testRuns removed

data TestPhase
    = TestsRunning [TestRun]            -- mixed running/completed during the run
    | TestsDone    [TestResult]

data PostBuild = PostBuild
    { testPhase :: TestPhase
    -- , evalResults :: …                 -- eval comments go here?
    }
```

What do you think?

Gipphe:
I like the way you think. Also makes it clear when something uses just the test results, eval comments results or the build result itself.

I'll implement your suggestion and see what happens with the `PostBuildStore` effect. Might be that it's redundant.

Gipphe:
I adapted it a bit from what you have there, but this is the gist of what I'm working towards:

Each test suite individually can have the state "running", "errored" or "completed". Thus, the `PostBuild` just holds a `Map TestTarget TestRun`.

I'll be making the same adaptations for eval comments down the line.

PR #267, opened 2026-07-09
feat: add tricorder log --path flag

cgeorgii:
- Print the path to the daemon log file instead of its contents
- Model the log command's mode as `LogMode` so `--path` and `--follow` are mutually exclusive at the parser level
- Document the `log` command and its flags in the tricorder SKILL


Gipphe (approved):
This is pretty good as-is, but you don't get a warning or error when passing both flags at the same time. `tricorder log --follow --path` results in `ShowLogPath` since `--path` is prioritized before `--follow` in the `logParser` (in that `a <|> b = a` as long as `a` is not `empty`).

I would also argue that the flag should be named `--print-path` instead of just `--path`, since `--path` sounds more like a flag that takes the path to a file or directory.

PR #519, opened 2026-09-17
reduce spurious wakeups

agentm:
tricorder was waking up every one to two seconds on threads which do not require it, causing GC to run. After this patch, the wakeups are gone.


Gipphe:
I agree with the changes to `FileWatcher`, but I'm gonna have to come back to you on Monday about the `IdleTimer` changes. I feel like there should be a better solution than "check every `maxCheckInterval` seconds" :sweat_smile: I agree that "check every 2 seconds", the current implementation, is no better though.

Gipphe (approved):
LGTM 👍 

I'll merge this now and deal with the change log in a separate PR. Thanks for the contribution! 💪

